PR #19, opened 2019-06-11
11.0 mail integration

aitorbouzas:
Three new methods:

`message_new` parses the mail and creates default values for the ticket
`message_update` subscribes partners discussing in the thread when a new mail is received
`message_get_suggested_recipients` checks whether the ticket has partner, if so it suggests to add it as a follower (if it is not added already). If the ticket hasn't a partner, it will check if it has a `partner_email` and add it to suggested recipients.


aitorbouzas:
Again! I didn't wanted to open a PR here... Sorry!

PR #26, opened 2019-07-01
11.0 ticket copy method

DaniKL:
[ADD] Ticket copy method

[ADD] Contributors


pedrobaeza (changes requested):
You shouldn't put 45 commits in the PR. Please apply the corresponding patch over the new module name.

PR #29, opened 2019-07-03
[FIX] Multi company

alvaro-domatix:
1.- Added company_id field (empty) to the initial creation of stages (data folder). This allows to use the initial stages with multi-company, so they are not created with any company.

2.- Added company_id field to the sequence. The purpose of this is to make posible having different sequences in each company.

3.- Refactored the "create" method to make posible different sequences.

4.- Deleted "readonly" attribute of company_id in the helpdesk.ticket form. (As it is in every other Odoo module.)

5.- Added a security rule to make posible to Users to see all tickets. (There is other group "User: Personal tickets" which won't have this permission.)

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -28,6 +28,7 @@
       <field name="code">helpdesk.ticket.sequence</field>
       <field name="prefix">HT</field>
       <field name="padding">5</field>
+      <field name="company_id"/>
     </record>
     <!-- Stages -->
     <record id="helpdesk_ticket_stage_new" model="helpdesk.ticket.stage">
@@ -35,18 +36,21 @@
       <field name="name">New</field>
       <field name="unattended">True</field>
       <field name="closed">False</field>
+      <field name="company_id"/>
     </record>
     <record id="helpdesk_ticket_stage_in_progress" model="helpdesk.ticket.stage">
       <field name="sequence">2</field>
       <field name="name">In Progress</field>
       <field name="unattended">False</field>
       <field name="closed">False</field>
+      <field name="company_id"/>
     </record>
     <record id="helpdesk_ticket_stage_awaiting" model="helpdesk.ticket.stage">
       <field name="sequence">3</field>
       <field name="name">Awaiting</field>
       <field name="unattended">False</field>
       <field name="closed">False</field>
+      <field name="company_id"/>
     </record>
     <record id="helpdesk_ticket_stage_done" model="helpdesk.ticket.stage">
       <field name="sequence">3</field>
@@ -54,6 +58,7 @@
       <field name="unattended">False</field>
       <field name="closed">True</field>
       <field name="fold">True</field>
+      <field name="company_id"/>
     </record>
     <record id="helpdesk_ticket_stage_cancelled" model="helpdesk.ticket.stage">
       <field name="sequence">4</field>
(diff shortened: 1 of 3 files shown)


aitorbouzas (approved):
Functional tested on runbot.
Code LGTM.

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -28,39 +28,45 @@
       <field name="code">helpdesk.ticket.sequence</field>
       <field name="prefix">HT</field>
       <field name="padding">5</field>
+      <field name="company_id" eval="False"/>
```
pedrobaeza: Why not `<field name="company_id"/>` in all for saving characters?

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -111,9 +111,15 @@ def _onchange_dominion_user_id(self):
     @api.model
     def create(self, vals):
         if vals.get('number', '/') == '/':
-            vals['number'] = self.env['ir.sequence'].next_by_code(
-                'helpdesk.ticket.sequence'
-            ) or '/'
+            if 'company_id' in vals:
```
pedrobaeza, line 115: A slight better code:
```python
seq = self.env['ir.sequence']
if 'company_id' in vals:
    seq = seq.with_context(force_company=vals['company_id'])
vals['number'] = seq.next_by_code('helpdesk.ticket.sequence') or '/'
```

helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -32,6 +32,12 @@
         <field name="domain_force">['|',('user_id','=',user.id),('user_id','=',False)]</field>
         <field name="groups" eval="[(4, ref('group_helpdesk_user_own'))]"/>
     </record>
+    <record id="helpdesk_all_tickets_rule" model="ir.rule">
```
pedrobaeza: You are introducing here other features not related to multi-company. Please remove this and the other one for being discussed in other PR.

pedrobaeza (changes requested):

helpdesk_mgmt/data/helpdesk_data.xml
alvaro-domatix: I'm not completely sure if I understand what you mean:

In the "ir.sequence" record I have added 'eval="False"' because it is how I have seen that Odoo does it in their code. Example:
https://github.com/odoo/odoo/blob/a3dcde21d51767d7ca9ab85da10e25ef24294248/addons/account/data/account_data.xml#L54

The rest of records are done the way I understand it has to be done, as well as you are saying, don't you?

alvaro-domatix:
Updated requested changes (security and sequence).

Waiting for feedback about the data records.

helpdesk_mgmt/data/helpdesk_data.xml
pedrobaeza: Yes, all works the same way. My proposal is for 3 things:

* Homogeneity for all the assignments. You are using 2 techniques.
* Clearer syntax: IMO this seems clearer (but it's subjective).
* Save code characters. Digital kittens will thank you :stuck_out_tongue_closed_eyes:

pedrobaeza: Do you agree then?

alvaro-domatix: Yes, Im changing from

`<field name="company_id" eval="False"/>` 

to 

`<field name="company_id"/>`

right now.

pedrobaeza: And the other occurrences of `<field name="company_id"></field>` to `<field name="company_id"/>`, please. (that's what I mean from the beginning)

pedrobaeza:
Please comment back when done.

PR #30, opened 2019-07-04
[ADD] Survey

samuel-domatix:
- New module that allows to send a survey to rate the support.


pedrobaeza:
Why not using rating module the same way is now integrated in most of the core modules (like project)? It can be an extra module (`helpdesk_mgmt_rating`), but reusing all that stuff.

PR #31, opened 2019-07-04
[ADD] Email templates for closed tickets and changed state.

miguel-domatix:
send an email if the ticket reaches a stage that has the associated template.

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -1,0 +1,248 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+  <data>
+    <record model="ir.module.category" id="module_helpdesk_category">
+      <field name="name">Helpdesk</field>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I miss:

* Use standard mail notification template as base for better design.
* Associate the template to the demo/data stages

miguel-domatix:
https://github.com/odoo/odoo/blob/11.0/addons/sale/data/mail_template_data.xml

A template like this?

pedrobaeza:
What about this other one that seems more beautiful (but this is an opinion)?

https://github.com/odoo/odoo/blob/1a62c2ed475d74c30222c30ec79aa4bdcdbca184/addons/sale/data/mail_template_data.xml#L65

About translations, the problem is that they can conflict as it has happened, so better to not translate in the same PR, but do it on Weblate.

pedrobaeza:
You have committed several merges, provoking conflicts

miguel-domatix:
We are closing this PR and opening another one in order to make the PR in the correct way.
Sorry for the inconvenience.

PR #32, opened 2019-07-04
11.0 attachments correction

AlhMaria:
The field of attachments now works correctly.

image


pedrobaeza:
Not that this page won't be needed in v12, as it's included in core. Please put a note about that in the code for easing migration.

PR #33, opened 2019-07-04
[ADD] Ticket creation from Portal

alvaro-domatix:
1.- Implemented 2 methods for managing the ticket creation from Portal.
2.- Creation of a form template to create tickets from Portal.
3.- Modified "portal_my_home" template to add a "Create Ticket" button if the user has not created any ticket yet. Otherwise, the button will only appear in "My tickets".
4.- Only users with Portal permission will be able to use the "Create Ticket" button.
5.- Closed tickets won't appear in "my tickets list" instead of tickets with closed_date.
6.- Some general fixes as security permissions, dependencies of old module name "helpdesk", groups for some menus...

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -1,6 +1,8 @@
 import werkzeug
 import logging
 import odoo.http as http
+import base64
+from openerp.http import request
 _logger = logging.getLogger(__name__)
 
 
@@ -12,8 +14,52 @@ class HelpdeskTicketController(http.Controller):
         values = {}
         for field_name, field_value in kw.items():
             values[field_name] = field_value
-        ticket = http.request.env['helpdesk.ticket'].sudo().\
-            search([('id', '=', values['ticket_id'])])
-        ticket.stage_id = values.get('stage_id')
-
+        ticket = http.request.env['helpdesk.ticket'].sudo().browse([(
+            'id', '=', values['ticket_id'])])
+        stage_id = int(values.get('stage_id'))
+        ticket.update({'stage_id': stage_id})
         return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
+
+    @http.route('/new/ticket', type="http", auth="user", website=True)
+    def create_new_ticket(self, **kw):
+        categories = http.request.env['helpdesk.ticket.category']. \
+            search([('active', '=', True)])
+        email = http.request.env.user.email
+        name = http.request.env.user.name
+        return http.request.render('helpdesk_mgmt.portal_create_ticket', {
+            'categories': categories, 'email': email, 'name': name})
+
+    @http.route('/submitted/ticket',
+                type="http", auth="user", website=True, csrf=True)
+    def submit_ticket(self, **kw):
+        vals = {
+            'partner_name': request.env.user.name,
+            'company_id': request.env.user.company_id.id,
+            'category_id': request.params.get('category'),
+            'partner_email': request.env.user.email,
+            'description': request.params.get('description'),
+            'name': request.params.get('subject'),
+            'attachment_ids': False,
+            'channel_id':
+                request.env['helpdesk.ticket.channel'].
+                sudo().search([('name', '=', 'Web')]).id,
+            'partner_id':
+                request.env['res.partner'].sudo().search([
+                    ('name', '=', request.env.user.name),
+                    ('email', '=', request.env.user.email)]).id
+        }
+        new_ticket = request.env['helpdesk.ticket'].sudo().create(
+            vals)
+        new_ticket.message_subscribe_users(user_ids=request.env.user.id)
+        if request.params.get('attachment'):
+            for c_file in request.httprequest.files.getlist('attachment'):
+                data = c_file.read()
+                if c_file.filename:
+                    request.env['ir.attachment'].sudo().create({
+                        'name': c_file.filename,
+                        'datas': base64.b64encode(data),
+                        'datas_fname': c_file.filename,
+                        'res_model': 'helpdesk.ticket',
+                        'res_id': new_ticket.id
+                    })
+        return werkzeug.utils.redirect("/my/tickets")
--- a/helpdesk_mgmt/security/ir.model.access.csv
+++ b/helpdesk_mgmt/security/ir.model.access.csv
@@ -5,7 +5,7 @@ access_helpdesk_ticket_user_personal,helpdesk.ticket.user.personal,model_helpdes
 access_helpdesk_ticket_portal,helpdesk.ticket.portal,model_helpdesk_ticket,base.group_portal,1,0,0,0
 access_helpdesk_ticket_stage_manager,helpdesk.ticket.stage.manager,model_helpdesk_ticket_stage,group_helpdesk_manager,1,1,1,1
 access_helpdesk_ticket_stage_user,helpdesk.ticket.stage.user,model_helpdesk_ticket_stage,base.group_user,1,0,0,0
-access_helpdesk_ticket_stage_portal,helpdesk.ticket.stage.portal,model_helpdesk_ticket_stage,base.group_portal,1,1,0,0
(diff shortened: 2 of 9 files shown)


helpdesk_mgmt/security/ir.model.access.csv
```diff
@@ -2,10 +2,10 @@ id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_helpdesk_ticket_manager,helpdesk.ticket.manager,model_helpdesk_ticket,group_helpdesk_manager,1,1,1,1
 access_helpdesk_ticket_user,helpdesk.ticket.user,model_helpdesk_ticket,group_helpdesk_user,1,1,1,0
 access_helpdesk_ticket_user_personal,helpdesk.ticket.user.personal,model_helpdesk_ticket,group_helpdesk_user_own,1,1,1,0
-access_helpdesk_ticket_portal,helpdesk.ticket.portal,model_helpdesk_ticket,base.group_portal,1,0,0,0
+access_helpdesk_ticket_portal,helpdesk.ticket.portal,model_helpdesk_ticket,base.group_portal,1,1,1,0
```
pedrobaeza: Don't do this! You are giving access to any ticket for being modified to a portal user for example using XML-RPC. Put the controller with sudo when creating the ticket instead.

alvaro-domatix: I fixed already your request change. Sorry for the delay.
I will be waiting for any other changes required.

Thank you Pedro.

helpdesk_mgmt/controllers/main.py
```diff
@@ -12,8 +14,64 @@ def support_ticket_close(self, **kw):
         values = {}
         for field_name, field_value in kw.items():
             values[field_name] = field_value
-        ticket = http.request.env['helpdesk.ticket'].sudo().\
-            search([('id', '=', values['ticket_id'])])
-        ticket.stage_id = values.get('stage_id')
+        ticket = http.request.env['helpdesk.ticket'].search([(
```
pedrobaeza: Do a `browse` instead and with sudo

helpdesk_mgmt/controllers/main.py
```diff
@@ -12,8 +14,64 @@ def support_ticket_close(self, **kw):
...
+            'id', '=', values['ticket_id'])])
+        stage_id = int(values.get('stage_id'))
+        ticket.sudo().update(
```
pedrobaeza: `ticket.stage_id = ...` after putting the sudo before

helpdesk_mgmt/controllers/main.py
```diff
@@ -12,8 +14,64 @@ def support_ticket_close(self, **kw):
...
 
         return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
+
+    @http.route('/new/ticket', type="http", auth="user", website=True)
+    def create_new_ticket(self, **kw):
+        categories = http.request.env['helpdesk.ticket.category']. \
+            search([('active', '=', True)])
+        email = http.request.env.user.email
+        name = http.request.env.user.name
+
```
pedrobaeza: Why leaving this meaningless empty lines inside the method? PEP8 discourages to do that.

helpdesk_mgmt/controllers/main.py
```diff
@@ -12,8 +14,64 @@ def support_ticket_close(self, **kw):
...
+        return http.request.render('helpdesk_mgmt.portal_create_ticket', {
+            'categories': categories, 'email': email, 'name': name})
+
+    @http.route('/submitted/ticket',
+                type="http", auth="user", website=True, csrf=True)
+    def submit_ticket(self, **kw):
+        vals = {'company_id': request.env.user.company_id.id}
+
+        for field_name, field_value in kw.items():
```
pedrobaeza: This is an open door for injecting ticket values. Can you limit this? How is done in other places in Odoo?

helpdesk_mgmt/views/helpdesk_ticket_templates.xml
```diff
@@ -17,18 +17,29 @@
                   <span class="badge" t-esc="ticket_count"/>
                   <a href="/my/tickets">Tickets</a>
               </li>
+              <t t-if="ticket_count == 0">
```
pedrobaeza: Shouldn't you show the new ticket in any case?

pedrobaeza (changes requested):

helpdesk_mgmt/controllers/main.py
samuel-domatix: Sorry but I don't understand what you mean.

pedrobaeza: That you don't need to sudo again. `ticket` is already "sudoed" thanks to previous browse with sudo.

helpdesk_mgmt/controllers/main.py
```diff
@@ -12,8 +14,53 @@ def support_ticket_close(self, **kw):
...
+            'attachment_ids': False,
+            'channel_id':
+                request.env['helpdesk.ticket.channel'].
+                sudo().search([('name', '=', 'Web')]).id,
+            'partner_id':
+                request.env['res.partner'].sudo().search([
+                    ('name', '=', request.env.user.name),
+                    ('email', '=', kw['email'])]).id
+        }
+        new_ticket_id = request.env['helpdesk.ticket'].sudo().create(
```
pedrobaeza: This is a record, not an ID, so call the variable `new_ticket`

helpdesk_mgmt/controllers/main.py
```diff
@@ -12,8 +14,53 @@ def support_ticket_close(self, **kw):
...
+        categories = http.request.env['helpdesk.ticket.category']. \
+            search([('active', '=', True)])
+        email = http.request.env.user.email
+        name = http.request.env.user.name
+        return http.request.render('helpdesk_mgmt.portal_create_ticket', {
+            'categories': categories, 'email': email, 'name': name})
+
+    @http.route('/submitted/ticket',
+                type="http", auth="user", website=True, csrf=True)
+    def submit_ticket(self, **kw):
```
pedrobaeza, line 34: Please don't use `**kw`, as this is a security hole, allowing to overwrite any field. Limit the arguments to the ones you want to parse.

samuel-domatix: Is it correct to do it with `request.params`? Although I have seen in the Odoo documentation that this is not usually used because `kw` provides you all the data.

pedrobaeza: OK, seeing Odoo code, they use the same technique, although they call the variable as `**post`.

pedrobaeza:
This pull request is containing a lot of changes, and thus the conflict has arisen. Please only one change set per PR.

alvaro-domatix:
Despite of this pull requests contains a lot of changes, they are almost all related with the Ticket creation. That's the reason why we made it into one pull request.
However, we will make as much pull requests as you consider. We would be grateful if you could tell us how could we split the code between several pull requests.
Thank you.

pedrobaeza:
Saying that is all related to "ticket creation" is the same as saying that all is related to the `helpdesk_mgmt`...

You have to split by set of logical changes, similar to the points you put on the main comment.

As you have done only one commit and all mixed, now you have to repeat the changes in a clean branch, observing the diff, there's no other way I'm afraid.

pedrobaeza:
Why asking to split? For several reasons:

- Being easier to be reviewed. If not, potential reviewers will discard huge diffs to review and this PR will hang up here for a lot.
- Being able to discuss separately each of the changes. There can be a conflicting one that blocks the rest that are OK.
- Avoid conflicts like the one you have, as others have already proposed similar patches for the specific problem.

alvaro-domatix:
I totally agree spliting code into more than one pull request. I actually understand the purpose of that.
What I am trying to say is that we are not sure about which could be the best way to split it, so I'm asking for a recomendation.
Anyway, we are glad you are taking your time to specify us the correct way to work with pull requests.
Thanks.

pedrobaeza:
As said, why don't you split them by each of the 6 points you put on the main comment? One of them is already addressed in another PR.

PR #34, opened 2019-07-04
[UPD] Attachment domain update

AlhMaria:
The field of attachments now works correctly.

image


pedrobaeza (approved):
This can be fast-tracked.

PR #40, opened 2019-07-15
[FIX]: Helpdesk TYPO Fixed in DEMO Data

agent-z28:


helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -26,7 +26,7 @@
 <data noupdate="1">
 
 
-    <record id="helpesk_ticket_personal_rule" model="ir.rule">
+    <record id="helpdesk_ticket_personal_rule" model="ir.rule">
```
pedrobaeza: You can't rename a noupdate="1" record, or you will have 2 records when updating. We have to stick with this bad name (but not really a problem), or do a migration script renaming the record.

helpdesk_mgmt/demo/helpdesk_demo.xml
```diff
@@ -32,7 +32,7 @@
     </record>
 
     <record id="helpdesk_team_2" model="helpdesk.ticket.team">
-        <field eval="&quot;Helpesk&quot;" name="name"/>
+        <field eval="&quot;Helpdesk&quot;" name="name"/>
```
pedrobaeza: This can be simplified as:
```suggestion
        <field name="name">Helpdesk</field>
```

helpdesk_mgmt/security/helpdesk_security.xml
agent-z28: Thats true:) i removed the change

helpdesk_mgmt/demo/helpdesk_demo.xml
agent-z28: The whole file has the &quot; i will not change it.
Just the typo

pedrobaeza: But you are changing that line, so fix it entirely. Other thing is that you don't touch that line.

dreispt (approved):

bodedra (approved):

dalonsod:
Hello, I've included this change directly in the v14 migration #179

pedrobaeza:
I have fixed it directly at https://github.com/OCA/helpdesk/commit/a496b94259ffa2d81b2298ab83517ce25f30e38d

PR #44, opened 2019-07-25
[UPD] Permissions Fixes

AlhMaria:
When you try to assing a ticket to another user, even being from the Helpdesk Manager team, a permission error occurs.

Here's a possible solution that I found:


pedrobaeza:
This is not correct. Maybe a record rule [(1, =, 1)] lacks for the group group_helpdesk_user

PR #47, opened 2019-07-29
Minor Fixes

samuel-domatix:
- Added default filter "Open tickets" in the tree view of tickets.
- Editing and creation from tree view in the views of channels, categories and tags.
- Added new fields in the tree views of channels, categories, stages, equipment and tags.
- Module logo update.
- Translations.


pedrobaeza:
Please, @SamuelDomatix, only one change per PR, and thus we can speed up those that are trivial, and let other with possible debate in hold. Any way, this branch is dirty with merge commits and both runbot and Travis totally red.

samuel-domatix:
Hi @pedrobaeza, how many PRs do you consider I should do with these changes?

pedrobaeza:
One per change. And translations are not done here, but on Weblate. So it seems 4.

emagdalenaC2i:
I think this PR can be closed

PR #49, opened 2019-07-29
Fix assigned date on ticket creation

samuel-domatix:
When creating a ticket and assigning a user, the assignment date is now updated correctly.


pedrobaeza:
Please describe the use case for this. Shouldn't this be done in other place?

Any way, you must check if there's already that val in the dict before overwriting it, and use direct assignment for smaller expression: `res['assigned_date'] = fields.Datetime.now()`

samuel-domatix:
When creating a new ticket and assigning a user, the _"assigned_date"_ field was not set. Therefore, if you have an assigned user at the time of creating it, that field should be modified.

I will make the changes you propose.

Thank you Pedro.

PR #50, opened 2019-07-29
Added default filter to ticket tree view

samuel-domatix:
- Added a default "Open tickets" filter in the tickets tree view.


helpdesk_mgmt/views/helpdesk_ticket_menu.xml
```diff
@@ -17,6 +17,9 @@
             <field name="res_model">helpdesk.ticket</field>
             <field name="view_mode">tree,kanban,form,pivot</field>
             <field name="view_type">form</field>
+            <field name="context">
+              {'search_default_open': 1}
```
pedrobaeza, line 21: Put all in one line for avoiding unneeded paddings.

samuel-domatix: I will fix it.

PR #51, opened 2019-07-29
Module logo update

samuel-domatix:
- Module logo changed


pedrobaeza:
Can you provide as well the source file (SVG or similar)? Shouldn't we differentiate from enterprise icon with another FA icon?

nacho-domatix:
Hi Pedro, we believe the icon is "different enough". As people say in Thailand, "same, same, but different" ;)
Anyway, is not going to create confusion at all cause even would be possible to use both (enterprise and OCA approach), I don't think that's gonna happen.

samuel-domatix:
This is the icon SVG file.

[helpdesk_icon.zip](https://github.com/OCA/helpdesk/files/3445709/helpdesk_icon.zip)

pedrobaeza:
Please include it in the PR in the same folder than the PNG.

pedrobaeza (approved):
Squashing on merge

emagdalenaC2i (approved):

Issue #55, opened 2019-07-31
Server date is used for initial value of last_stag_update


luffah:
In ticket list view `web#model=helpdesk.ticket&view_type=list` : I have ~~inconsistency~~
something i did not expect on last stage update value (last update is before creation...) : the default last stage update is the server start time.

Ticket number | Title | Partner Name | Assigned user | Stage | Priority | Created on | Last Stage Update
 -- | -- | -- | -- | -- | -- | -- | --
HT00032 | Rechargement vue |   |   | New |   | 30/07/2019 12:07:14 | 30/07/2019 11:39:04

Solution in `models/helpdesk_ticket.py`, remove parenthesis in Datetime.now()
```python
    last_stage_update = fields.Datetime(
        string='Last Stage Update',
        default=fields.Datetime.now,
    )
```

pedrobaeza:
Yes, indeed. Please open a PR.

PR #56, opened 2019-08-01
[IMP] helpdesk_mgmt: at creation last_stage_update use actual now

luffah:
In helpdesk ticket, default last_stage_update use actual now instead of server start time.
More details in #55

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -43,7 +43,7 @@ class HelpdeskTicket(models.Model):
 
     last_stage_update = fields.Datetime(
         string='Last Stage Update',
-        default=fields.Datetime.now(),
+        default=fields.Datetime.now,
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you check Travis fail?

luffah:
> Can you check Travis fail?

Maybe the test is too much "strict" .. it does test if the last_stage_update before `ticket.write({'stage_id': self.stage_closed.id})` is strictly older that the one after the write, although the time is registered according to the DEFAULT_SERVER_DATETIME_FORMAT which have the second as unit. (2 `assert` + 1 `write` instructions lasts less than 1 second)

How to run the test on my container ? I want to verify it the error occur when i add a `time.sleep(1)` between the 2 instructions.

pedrobaeza:
Add that instruction here in the PR and Travis will tell without needing to deploy it locally.

emagdalenaC2i (approved):

PR #59, opened 2019-08-01
Imp] helpdesk mgmt at creation last stage update use actual now

emagdalenaC2i:
See #56


pedrobaeza:
Please cherry-pick commit instead for keeping attribution.

PR #62, opened 2019-08-06
[IMP] helpdesk_mgmt: at creation last_stage_update use actual now

luffah:
In helpdesk ticket, default last_stage_update use actual now instead of server start time.
More details in https://github.com/OCA/helpdesk/issues/55

This PR propagate to v12 change made by https://github.com/OCA/helpdesk/pull/56


pedrobaeza (approved):
Fast-tracking

PR #66, opened 2019-08-07
[ADD] Email templates for closed tickets and changed state

miguel-domatix:
Email templates to use when the ticket reaches a stage that has the associated template.

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -22,6 +22,169 @@
         <p>The ticket ${object.number} has been assigned to you.</p>
       </field>
     </record>
+
+    <record id="closed_ticket_template" model="mail.template">
+      <field name="name">Helpdesk Closed Ticket Notification Email</field>
+      <field name="model_id" ref="helpdesk_mgmt.model_helpdesk_ticket"/>
+      <field name="email_from">${object.company_id.partner_id.email}</field>
+      <field name="subject">The ticket ${object.number} has been closed.</field>
+      <field name="partner_to">${object.partner_id.id}</field>
+      <field name="auto_delete" eval="False"/>
+      <field name="lang">${object.partner_id.lang}</field>
+      <field name="body_html"><![CDATA[<html>
+                <head></head>
+                % set record = ctx.get('record')
+                % set company = record and record.company_id or ctx.get('company')
+                <body style="margin: 0; padding: 0;">
+                <table border="0" width="100%" cellpadding="0" bgcolor="#ededed" style="padding: 20px; background-color: #ededed; border-collapse:separate;" summary="o_mail_notification">
+                    <tbody>
+                      <!-- HEADER -->
+                      <tr>
+                        <td align="center" style="min-width: 590px;">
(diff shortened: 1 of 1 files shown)


helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -22,6 +22,169 @@
...
+                                    </a>
+                                % endif
+                              </td>
+                            </tr>
+                          </table>
+                        </td>
+                      </tr>
+                      <tr>
+                        <td align="center">
+                            Powered by <a target="_blank" href="https://www.odoo.com">Odoo</a>.
```
pedrobaeza: Why adding this text?

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -22,6 +22,169 @@
...
```
pedrobaeza: Why adding this text?

pedrobaeza (commented):
Not sure that the stage change should be notified to the customer. And how is this sent? This is only adding the templates, not any automation code.

miguel-domatix:
This PR is only for template emails. Furthermore, each stage can have a email template that notify. Then, if in this PR must be the actions for templates, tell me.
Thank you.

pedrobaeza:
Well, I think the added value is to have them usable installing the module without extra steps.

miguel-domatix:
Do I have to associate the template with the corresponding states from the data folder?

pedrobaeza:
I don't know right now how do you get the mail sent. Is there such automation for auto-sending a mail template when the stage is changed like the one in tasks? If so, then linking it with the stage in data is enough.

miguel-domatix:
In helpdesk_ticket.py file, there are a function that send a email if the state changes if it has an associated template. The function's name is _track_template.

pedrobaeza:
Perfect, then just link that template to the standard stages populated in data

pedrobaeza:
Any way, my doubt about spamming user each time you change a stage is there.

miguel-domatix:
In the states form you can configure it and associate or not any email template.

pedrobaeza:
Yeah, then it's better to not link that template by default.

miguel-domatix:
Fine, then I let it like this.

pedrobaeza:
But this is not linking the closed template... What I don't think it's good is to link the stage changed template, but ticket closed template should be.

miguel-domatix:
The email templates aren't shown on chatter, but it works on mail. Do I make a simpler template for chatter and these two for mail?

pedrobaeza:
Please look at project stages, where you have a mail template linked to them, and when you changed the stage, there's an automation that sends that template. That's the idea here, with the closed stage (provided as data with that template linked).

miguel-domatix:
Ok. Then, this templates is not seen in chatter. I think the chatter can't render the template. That may be?.

pedrobaeza:
Chatter shows any message sent in the thread, with or without template, with or without "advanced formatting". Please review task stages code for knowing how to send a message with a template via code.

miguel-domatix:
Did you mean do it like that?

pedrobaeza:
@miguel-domatix yes, finally you got me!

miguel-domatix:
Fine!

alvaro-domatix (approved):

samuel-domatix (approved):

PR #71, opened 2019-08-09
[ADD] Creation of a form template to create tickets from Portal.

samuel-domatix:
Added a creation form in the portal to enter the ticket data to be created.


pedrobaeza:
Maybe this should be tight to the other 2 PRs that are creating the interface and modifying the controller, as all of this is a whole feature (not like for example, permissions).

samuel-domatix:
Ok, I will make a PR with all the changes that involve the creation of tickets.

PR #79, opened 2019-09-02
Linking tickets to other documents

samuel-domatix:
- Added helpdesk settings.
- Added a new menu to access the settings.
- Added two fields in settings to enable or disable this option and select the models to relate.
- Added new field in the ticket to relate this with other documents.

### DIFF
--- a/helpdesk_mgmt/__manifest__.py
+++ b/helpdesk_mgmt/__manifest__.py
@@ -32,6 +32,7 @@
         'views/helpdesk_ticket_tag_view.xml',
         'views/helpdesk_ticket_view.xml',
         'views/helpdesk_dashboard_view.xml',
+        'views/res_config_settings.xml',
     ],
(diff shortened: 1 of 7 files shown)


alvaro-domatix (approved):

miguel-domatix (approved):

emagdalenaC2i:
First, I am not sure this should be part of the base heldesk_mgmt module or an extra module. @aitorbouzas and @pedrobaeza what do you think?

Second, you can see the two new fields when you edit an existing ticket, but you don't see it when you create a new one.

pedrobaeza:
Well, the feature is not very different from `crm_claim` module that includes a reference field, but it's a bit limited, as you can only link one document. Another future problem is that referable models dissappears in v12.

samuel-domatix:
Ok, I'll do it in a separate module.

PR #81, opened 2019-09-06
11.0 link documents

miguel-domatix:
- Added helpdesk settings.
- Added a new menu to access the settings.
- Added two fields in settings to enable or disable this option and select the models to relate.
- Added new field in the ticket to relate this with other documents.


helpdesk_mgmt_link_documents/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': 'Helpdesk linker',
+    'summary': 'Allow link any document to helpdesk tickets',
+    'version': '11.0.1.8.0',
```
emagdalenaC2i:
```suggestion
    'version': '11.0.1.0.0',
```

helpdesk_mgmt_link_documents/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': 'Helpdesk linker',
```
emagdalenaC2i:
```suggestion
    'name': 'Helpdesk link documents',
```

helpdesk_mgmt_link_documents/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'summary': 'Allow link any document to helpdesk tickets',
+    'version': '11.0.1.8.0',
+    'license': 'AGPL-3',
+    'author': 'Domatix',
+    'website': 'https://www.domatix.com/',
```
emagdalenaC2i:
```suggestion
    'website': 'https://github.com/OCA/helpdesk',
```

emagdalenaC2i (changes requested):
Please add the new README by fragment system copying https://github.com/OCA/maintainer-tools/tree/master/template/module/readme and editing or removing the needed sections

pedrobaeza:
Please don't use plurals in module name: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules

And I don't get why you finally did a module, as I said that it's OK to add a reference field in existing module.

miguel-domatix:
I have made a new module for this:
https://github.com/OCA/helpdesk/pull/79#issuecomment-528285794

pedrobaeza:
Well, I pointed several possible future problems. Any way, I see you have change the plural in the manifest, not the technical name (that's to say, the folder name).

nacho-domatix:
> Well, I pointed several possible future problems.

This feature might be interesting for many companies. We can add it as an option (so can be disabled by UI) or as it is now (inheritance module).

pedrobaeza:
I'm comfortable with both. What do you prefer?

nacho-domatix:
UI option indeed, it's more "Odoo" this way.

PR #88, opened 2019-09-13
[ADD] 12.0 helpdesk_reopen

PierrickBrun:
reopends ticket if a mail is received concerning it


helpdesk_reopen/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "name": "Helpdesk Reopen",
+    "version": "12.0.0.1.0",
+    "category": "Helpdesk",
+    "summary": "Auto Helpdesk Reopen",
+    "description": """
+	Automatically reopens helpdesk queries if someone replies back to that query
+	""",
+    "author": "OERP, Coop IT Easy SCRLfs",
+    "website": "www.oerp.eu",
+    "depends": ["helpdesk", "mail"],
```
rvalyi: helpdesk already depends on mail, so I think it's not necessary to add the dependency here again

rvalyi:
Hello @PierrickBrun. Glad to see you are contributing to the great helpdesk project. One question though: it seems you are not the author of the module, right? What is the context then? Do we have some permission from the original module author to host his module at the OCA? Even if the module is originally AGPL, there is also a copyright grant to the OCA to accept, so we really need that permission written somewhere...

helpdesk_reopen/__init__.py
```diff
@@ -0,0 +1,2 @@
+# coding: utf-8
```
daramousk: Encoding not needed in Python3, replace this with the Licence line

helpdesk_reopen/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Helpdesk Reopen",
+    "version": "12.0.0.1.0",
+    "category": "Helpdesk",
+    "summary": "Auto Helpdesk Reopen",
+    "description": """
+	Automatically reopens helpdesk queries if someone replies back to that query
+	""",
+    "author": "OERP, Coop IT Easy SCRLfs",
```
daramousk: Please go through $, the author key needs to contain OCA https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules

helpdesk_reopen/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+# coding: utf-8
```
daramousk: remove coding from everywhere

daramousk (changes requested):
Hey there! Thank you for this contribution, please create a readme file as instructed here https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py so that we have further information about this module and what it does. Then we can continue.

Also, perhaps we can avoid writing code by creating some sort of automated action running?

PR #95, opened 2019-09-28
[12.0] [FIX] Ticket creation from portal - message_subscribe_users > message_subscribe

marcelsavegnago:
Ticker portal creation

- Fix message_subscribe_user > message_subscribe


pedrobaeza:
It seems it still continues in red. Why do you want to do this change?

marcelsavegnago:
> It seems it still continues in red. Why do you want to do this change?

The message_subscribe_users function has apparently changed to message_subscribe. When I try to create a ticket via portal Odoo returns that this function does not exist in version 12.0.

About the error reported by Travis I do not know what it can be.

luffah:
I (as a low rank contributor) approve this change.

luffah:
Shall Ticket category be mandatory in portal while it is not in model ?

marcelsavegnago:
> It seems it still continues in red. Why do you want to do this change?

@pedrobaeza Can you guide me regarding the error occurred in Travis? I honestly do not understand what it can be :) sorry

luffah:
Since the PR is not validated yet. Why limit access to button "new ticket" to the restrictive group "portal user", given only these users could see the list  and portal home? (it remove the possibility to test as admin)

pedrobaeza:
The problem is:
```
helpdesk_mgmt/controllers/main.py:56:80: E501 line too long (81 > 79 characters)
```

pedrobaeza (approved):

marcelsavegnago:
> Since the PR is not validated yet. Why limit access to button "new ticket" to the restrictive group "portal user", given only these users could see the list and portal home? (it remove the possibility to test as admin)

@luffah I prefer to focus on fixing this problem first so that I can release the fix as soon as possible. What do you think about submitting a new PR? I think working with minor corrections speeds up the process. But, I may be wrong.

emagdalenaC2i (approved):

DarioLodeiros (approved):

PR #98, opened 2019-10-01
[NEW] Send email to the contact that create the ticket

mileo:
Signed-off-by: Luis Felipe Mileo <mileo@kmee.com.br>

When you use the complete email form, without creating the partner, the email was not filled with the partner_email only on the fast reply.

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -13,6 +13,7 @@
       <field name="name">Ticket Assignment</field>
       <field name="model_id" ref="model_helpdesk_ticket"/>
       <field name="email_from">${object.company_id.partner_id.email}</field>
+      <field name="email_cc">${not object.partner_id and object.partner_email or ''|safe},</field>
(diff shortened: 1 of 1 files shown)


helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -13,6 +13,7 @@
       <field name="name">Ticket Assignment</field>
       <field name="model_id" ref="model_helpdesk_ticket"/>
       <field name="email_from">${object.company_id.partner_id.email}</field>
+      <field name="email_cc">${object.partner_email|safe},</field>
```
pedrobaeza:
```suggestion
      <field name="email_cc">${not object.partner_id and object.partner_email or ''|safe},</field>
```
for avoiding double sending if partner is present

rafaelbn (approved):
:+1: 

@carlos-domatix @DarioLodeiros could you review please?

carlos-domatix (approved):

DarioLodeiros (approved):
:+1:

PR #104, opened 2019-10-17
13.0 mig helpdesk mgmt

emagdalenaC2i:


helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -139,7 +139,7 @@ def copy(self, default=None):
         return res
 
     def write(self, vals):
-        for ticket in self:
+        for _ticket in self:
```
DarioLodeiros, line 141: Why use a recordset loop if we are working only with vals parameter?

emagdalenaC2i: I do not know. It doesn't make sense, but it's like that for Odoo 11 and 12

cubells:
@emagdalenaC2i @DarioLodeiros 
Are you going to continue with this or do you want I finish this migration?

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -0,0 +1,223 @@
...
+      <field name="name">Email</field>
+    </record>
+    <record id="helpdesk_ticket_channel_phone" model="helpdesk.ticket.channel">
+      <field name="name">Phone</field>
+    </record>
+    <record id="helpdesk_ticket_channel_other" model="helpdesk.ticket.channel">
+      <field name="name">Other</field>
+    </record>
+    <!--default alias for leads-->
+    <record id="mail_alias_lead_info" model="mail.alias">
```
angelmoya, line 215: You can not use this id "mail{alias_lead_info" because it is used on crm module, and this do that module crm will be uninstallable.

cubells:
Supersede in https://github.com/OCA/helpdesk/pull/117

cubells:
@angelmoya 
can you test this new pr (https://github.com/OCA/helpdesk/pull/117)?

emagdalenaC2i:
@cubells I close this PR and will try to test #117 soon

PR #105, opened 2019-10-21
[12.0][ADD] alias mixin model to team

DarioLodeiros:
Add alias mixin model and standard logic to automated tickets creation from mail

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket_team.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket_team.py
@@ -5,6 +5,7 @@ class HelpdeskTeam(models.Model):
 
     _name = 'helpdesk.ticket.team'
     _description = 'Helpdesk Ticket Team'
+    _inherit = ['mail.thread', 'mail.alias.mixin']
 
     name = fields.Char(string='Name', required=True)
     user_ids = fields.Many2many(comodel_name='res.users', string='Members')
@@ -18,7 +19,10 @@ class HelpdeskTeam(models.Model):
         default=lambda self: self.env['res.company']._company_default_get(
             'helpdesk.ticket')
     )
-
+    alias_id = fields.Many2one(help="The email address associated with "
+                               "this channel. New emails received will "
+                               "automatically create new tickets assigned "
+                               "to the channel.")
     color = fields.Integer("Color Index", default=0)
 
     ticket_ids = fields.One2many(
@@ -62,3 +66,11 @@ class HelpdeskTeam(models.Model):
             record.todo_ticket_count_high_priority = len(
                 record.todo_ticket_ids.filtered(
                     lambda ticket: ticket.priority == '3'))
+
+    def get_alias_model_name(self, vals):
+        return 'helpdesk.ticket'
+
+    def get_alias_values(self):
+        values = super(HelpdeskTeam, self).get_alias_values()
+        values['alias_defaults'] = {'team_id': self.id}
+        return values
--- a/helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
@@ -35,6 +35,23 @@
                         <field name="user_ids" widget="many2many_tags"/>
                         <field name="company_id" groups="base.group_multi_company"/>
                     </group>
+                    <group>
+                        <group name="left">
+                            <field name="active" invisible="1"/>
+                            <label for="alias_name" string="Email Alias"/>
+                            <div class="oe_inline" name="alias_def">
+                                <field name="alias_id" class="oe_read_only oe_inline"
+                                    string="Email Alias" required="0"/>
+                                <div class="oe_edit_only oe_inline" name="edit_alias" style="display: inline;" >
+                                    <field name="alias_name" class="oe_inline"/>@<field name="alias_domain" class="oe_inline" readonly="1"/>
+                                    <button  icon="fa-arrow-right" type="action" name="%(base_setup.action_general_configuration)d" string="Configure domain name" class="btn-link" attrs="{'invisible': [('alias_domain', '!=', False)]}"/>
+                                </div>
+                            </div>
+                            <field name="alias_contact" class="oe_inline"
+                                string="Accept Emails From" />
+                            <field name="alias_user_id" string="Assign to"/>
+                        </group>
+                    </group>
                 </sheet>
             </form>
         </field>


DarioLodeiros:
#104 for v12
and https://github.com/Change2improve/helpdesk/pull/1/commits/0b78e78e950f808dcf79e7e0adaf7f3826417807 for v13

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -211,5 +211,13 @@
     <record id="helpdesk_ticket_channel_other" model="helpdesk.ticket.channel">
       <field name="name">Other</field>
     </record>
+    <!--default alias for leads-->
```
pedrobaeza: Leads? Why is this in data and not in demo? There shouldn't be any "default" alias, and more if empty.

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -18,7 +20,14 @@ class HelpdeskTeam(models.Model):
         default=lambda self: self.env['res.company']._company_default_get(
             'helpdesk.ticket')
     )
-
+    user_id = fields.Many2one('res.users', string='Team Leader',
+                              check_company=True)
```
pedrobaeza: Why `check_company=True`?

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -18,7 +20,14 @@ class HelpdeskTeam(models.Model):
         default=lambda self: self.env['res.company']._company_default_get(
             'helpdesk.ticket')
     )
-
+    user_id = fields.Many2one('res.users', string='Team Leader',
```
pedrobaeza: Why do you need to declare these fields if you are inheriting from `mail.alias.mixin`?

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -61,4 +70,15 @@ def _compute_todo_tickets(self):
...
-                    lambda ticket: ticket.priority == '3'))
+                    lambda ticket: ticket.priority == "3")
+            )
+
+    def get_alias_model_name(self, vals):
+        return 'helpdesk.ticket'
+
+    def get_alias_values(self):
+        values = super(HelpdeskTeam, self).get_alias_values()
+        values['alias_defaults'] = defaults = safe_eval(
```
pedrobaeza: Why do yo need to do this manually?

helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
```diff
@@ -31,11 +31,54 @@
...
+                            <field name="alias_contact" class="oe_inline"
+                                string="Accept Emails From" />
+                            <field name="alias_user_id" string="Assign to"/>
+                            <field name="company_id" options="{'no_create': True}" groups="base.group_multi_company"/>
+                        </group>
+                        <group name="right">
+                        </group>
                     </group>
+                    <notebook>
+                        <page name="members" string="Team Members" >
```
pedrobaeza: And how is this related to alias?

pedrobaeza (changes requested):

helpdesk_mgmt/data/helpdesk_data.xml
DarioLodeiros: You are right... this record not has sense (on CRM use it to compute a default alias for leads, but here we do not need anything similar)

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: Like the sale team reference, to make sure the team/user coherence company:
https://github.com/OCA/OCB/blob/13.0/addons/sales_team/models/crm_team.py#L51

pedrobaeza: I usually see this as a problem, as a user can belong to several companies and this prevents it if the user is logged in another one.

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: I can separate this change in other pr... it has not relation with alias

pedrobaeza: Yes, please. Each changeset per PR, so they can be discussed separately and one can be faster than the other.

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: "Return values to create an alias, or to write on the alias after its creation"... I think that it is the better option to get alias values when you create or write from the helpdesk team link model

pedrobaeza: But you don't have to parse existing default values... Just add what you want, but let super to do the rest. Or better, which is the core strategy, save proper defaults on alias creation.

helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
DarioLodeiros: Idem that user_ids... It is not related with alias... other PR, no? :)

pedrobaeza: Yes!

helpdesk_mgmt/data/helpdesk_data.xml
DarioLodeiros: Done

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: I'll leave the field for another pr, so we'll check it out

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: Done!

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: For the moment I take it off so I can do the merge, I would like to check the flow on sale team alias, and view if it makes sense in this case... 
https://github.com/OCA/OCB/blob/12.0/addons/crm/models/crm_team.py#L81
anyway we can see it in its pr.

helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
DarioLodeiros: Done

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -18,7 +19,12 @@ class HelpdeskTeam(models.Model):
         default=lambda self: self.env['res.company']._company_default_get(
             'helpdesk.ticket')
     )
-
+    alias_id = fields.Many2one('mail.alias', string='Alias',
```
pedrobaeza: mail.alias.mixin already defines this field:

https://github.com/odoo/odoo/blob/dd68d151426e971d393c173e43fd8a1f16ade294/addons/mail/models/mail_alias.py#L191

Why are you redefining it? If you want to add the help attribute, just put that one, not the rest. Also use a better layout for avoiding a lot of lines and no need to use `\`, but close on each line the quote and start a new one in the following, and they are concatenated (respect the space inside the quotes):
```python
    alias_id = fields.Many2one(
         help="The email address associated with "
              "this channel...
```

helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
```diff
@@ -35,6 +35,25 @@
...
+                                <div class="oe_edit_only oe_inline" name="edit_alias" style="display: inline;" >
+                                    <field name="alias_name" class="oe_inline"/>@<field name="alias_domain" class="oe_inline" readonly="1"/>
+                                    <button  icon="fa-arrow-right" type="action" name="%(base_setup.action_general_configuration)d" string="Configure domain name" class="btn-link" attrs="{'invisible': [('alias_domain', '!=', False)]}"/>
+                                </div>
+                            </div>
+                            <field name="alias_contact" class="oe_inline"
+                                string="Accept Emails From" />
+                            <field name="alias_user_id" string="Assign to"/>
+                        </group>
+                        <group name="right">
```
pedrobaeza: Why this extra empty group?

pedrobaeza (changes requested):
You still need to overwrite `def get_alias_values(self):` method, but no need to `safe_eval` anything. Just add in the returned dictionary the key for the defaults and the wanted defaults.

helpdesk_mgmt/models/helpdesk_ticket_team.py
DarioLodeiros: I think like you, but when checkin de code in the core, I can see this redefinition in all models... (hr_job, crm_team, project, maintenaince...)... anyway I prefer not to redefine attributes. Done :)

pedrobaeza: @DarioLodeiros this is not a question of thinking one way or the other, but how ORM works. Odoo is not a model for checking for better practices, as it contains a lot of not updated code to latest features of the ORM. Full definition is going to work, but this forces you to:

- Retranslate terms.
- If upstream definition is updated through versions, you have to adapt the redefinition to that changes.

helpdesk_mgmt/views/helpdesk_ticket_team_view.xml
DarioLodeiros: copy/pastle issue... :sob: 
https://github.com/OCA/OCB/blob/13.0/addons/sales_team/views/crm_team_views.xml#L77

helpdesk_mgmt/models/helpdesk_ticket_team.py
pedrobaeza: @DarioLodeiros this is not a question of thinking one way or the other, but how ORM works. Odoo is not a model for checking for better practices, as it contains a lot of not updated code to latest features of the ORM. Full definition is going to work, but this forces you to:

* Retranslate terms.
* If upstream definition is updated through versions, you have to adapt the redefinition to that changes.

DarioLodeiros: Of course... I just wanted to say that it was guided by the references found ;)
Thanks Pedro!

pedrobaeza (approved):
It was not so difficult, eh! :wink:

DarioLodeiros:
Thanks for the review!

elvise:
Any good for merger ? 😄

rafaelbn (approved):
:+1:

PR #106, opened 2019-10-24
[WIP][12.0] Base for website_helpdesk

ananiasfilho:
Allow portal users to create helpdesk tickets from website


pedrobaeza:
AFAIK, this is already done. @carlosDomatix can you confirm?

mileo:
> AFAIK, this is already done. @carlosDomatix can you confirm?

I remember @ericflaux  from @Osiell saying that someone was working on it during oca days. but could not find the pull request.

DarioLodeiros:
Its avail on the portal page:
https://github.com/OCA/helpdesk/blob/12.0/helpdesk_mgmt/controllers/main.py#L27
In this record you can view the original portal form ticket creation, and then, the option in this PR (I think that is better continue with the original option... ??)
HelpDeskTicketWebSitePortal

carlos-domatix:
> AFAIK, this is already done. @carlosDomatix can you confirm?

Yes, this is already done, as @DarioLodeiros has shown.

DarioLodeiros:
I have a doubt with this question on v13... its better to integrate de ticket creation form with the core form editor registry released on 13.0?
Example with recruitment:
https://github.com/OCA/OCB/blob/13.0/addons/website_hr_recruitment/static/src/js/website_hr_recruitment_editor.js#L9

pedrobaeza:
Why don't you put that on the migration issue?

PR #113, opened 2020-01-15
[ADD] Italian translation

mauriziobrilli:


pedrobaeza:
This shouldn't be done here, but on https://translation.odoo-community.org

Issue #120, opened 2020-04-27
[12.0] "Assigned to" email notification sent to partner


francesco-ooops:
When the ticket is assigned to a user, the email notification "Hello *user*,
The ticket HT00001 has been assigned to you." is sent to the partner (and not to the user)

This happens even if partner is not follower of the ticket.

Steps to reproduce: 

1) create new ticket
2) assign to user
3) check outgoing email in settings > tecnical > emails

PR #122, opened 2020-05-18
[FIX] Repeated active field failure in the helpdesk team view

cramosh:
In the team view form active's smartbutton did not show the string because it was repeated in the view and showed the button wrong

PR #125, opened 2020-05-18
#Issue 120 [12.0] "Assigned to" email notification sent to partner

cramosh:
Solved issue #120. 

When assigning the user a ticket, the email is reported to the user and not to the partner, selecting the user's language

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -13,11 +13,10 @@
       <field name="name">Ticket Assignment</field>
       <field name="model_id" ref="model_helpdesk_ticket"/>
       <field name="email_from">${object.company_id.partner_id.email}</field>
-      <field name="email_cc">${not object.partner_id and object.partner_email or ''|safe},</field>
       <field name="subject">${object.company_id.name} Ticket Assignment (Ref ${object.number or 'n/a' })</field>
-      <field name="partner_to">${object.partner_id.id}</field>
+      <field name="email_to">${object.user_id.partner_id.email}</field>
       <field name="auto_delete" eval="False"/>
-      <field name="lang">${object.partner_id.lang}</field>
+      <field name="lang">${object.user_id.partner_id.lang}</field>
       <field name="body_html" type="xml">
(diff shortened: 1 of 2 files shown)


emagdalenaC2i (approved):
LGTM Tested in runbot.
But rebase once #122 is merged, and squash commits

GorkaMagra (approved):
Technical review and tested in runbot with no issues whatsoever. Approved

pedrobaeza (changes requested):
The change is not valid because you are losing some data if always using a raw email. The strategy should be to use partner if available, and if not, then use raw email.

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -13,11 +13,10 @@
       <field name="name">Ticket Assignment</field>
       <field name="model_id" ref="model_helpdesk_ticket"/>
       <field name="email_from">${object.company_id.partner_id.email}</field>
-      <field name="email_cc">${not object.partner_id and object.partner_email or ''|safe},</field>
       <field name="subject">${object.company_id.name} Ticket Assignment (Ref ${object.number or 'n/a' })</field>
-      <field name="partner_to">${object.partner_id.id}</field>
+      <field name="email_to">${object.user_id.partner_id.email}</field>
```
DarioLodeiros, line 17: Change this to: `<field name="partner_to">${object.user_id.partner_id.id}</field>`

DarioLodeiros: The partner is the only way to send "a assigned to" mail (there is no separate mail field)

lukejcn: @DarioLodeiros Thank you, this appears to have stopped sending the customer an 'assigned to' email when a ticket is created from an inbound email. The odoo user doesn't get any notification though, nothing in 'discuss' (the users have 'handle notification in odoo' set). Is this the same for others?

** UPDATE **
I take it back. When a user emails into the team and a ticket gets created, they're automatically added as a follower. I'm assuming they shouldn't be made a follower but instead the assigned user should?

PR #134, opened 2020-06-11
[12.0][FIX] Use HTML field type for descriptions, to match widgets

janverb:
HTML fields do some additional processing that's good to have in this case.

If this looks good I'll see which other versions need it.


pedrobaeza (approved):
Thanks for the change!

Fast-tracking for having this needed patch

Issue #137, opened 2020-06-11
 [13.0] helpdesk_mgmt : portal users can't comment their own tickets


luffah:
Hello, 

   (at commit b2f7d3c48187c4aed7d79b59e9dbe4df6be9a8de)

   after successfully creating a ticket from portal from ticket with a portal user.
   This one can't comment.
   If you fill the message field, and send the message your redirected to this html error message.
```
Sorry, you are not allowed to modify documents of type 'Helpdesk Ticket' (helpdesk.ticket). This operation is allowed for the groups:
	- Helpdesk/Helpdesk Manager
	- Helpdesk/User
	- Helpdesk/User: Personal tickets - (Operation: write, User: 6)
```

luffah:
nice job :+1:

PR #139, opened 2020-06-16
[FIX] helpdesk_mgmt: add write permission for portal user

davejames:
Fixes issue #137 where where permission error was encountered when portal user tries to add a message to their ticket

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -7,6 +7,7 @@ class HelpdeskTicket(models.Model):
     _description = "Helpdesk Ticket"
     _rec_name = "number"
     _order = "number desc"
+    _mail_post_access = "read"
     _inherit = ["mail.thread.cc", "mail.activity.mixin"]
 
     def _get_default_stage_id(self):


pedrobaeza:
I don't think this is the solution, or at least not without an extra record rule.

davejames:
> I don't think this is the solution, or at least not without an extra record rule.

Where do you see the issue? There is already a record rule applicable to portal users

`        <record id="helpdesk_ticket_rule_portal" model="ir.rule">
            <field name="name">Portal Personal Tickets</field>
            <field name="model_id" ref="model_helpdesk_ticket" />
            <field
                name="domain_force"
            >[('message_partner_ids','child_of',[user.commercial_partner_id.id])]</field>
            <field name="groups" eval="[(4, ref('base.group_portal'))]" />
        </record>
`

pedrobaeza:
OK, if there is a record rule, then the access is limited. Any way, I would check for how it's done in other models like `sale.order` for letting users to write messages.

davejames:
@pedrobaeza you're right, I found a better solution

davejames:
@pedrobaeza @DarioLodeiros would you be able to review this? It's just a one-liner but it fixes the bug

pedrobaeza (approved):

DarioLodeiros (approved):

PR #144, opened 2020-07-08
[12.0] Improve helpdesk_mgmt

eLBati:
See commits for detailed descriptions

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -122,6 +122,12 @@ class HelpdeskTicket(models.Model):
             vals.setdefault("partner_name", partner.name)
             vals.setdefault("partner_email", partner.email)
 
+        if self.env.context.get(
+            'fetchmail_cron_running'
+        ) and not vals.get('channel_id'):
+            vals['channel_id'] = self.env.ref(
+                'helpdesk_mgmt.helpdesk_ticket_channel_email').id
+
         res = super().create(vals)
 
         # Check if mail to the user has to be sent
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -7,7 +7,7 @@ class HelpdeskTicket(models.Model):
     _name = 'helpdesk.ticket'
     _description = 'Helpdesk Ticket'
     _rec_name = 'number'
-    _order = 'number desc'
+    _order = 'priority desc, sequence, number desc'
     _inherit = ['mail.thread', 'mail.activity.mixin', 'portal.mixin']
 
     def _get_default_stage_id(self):
@@ -78,6 +78,9 @@ class HelpdeskTicket(models.Model):
         ('normal', 'Default'),
         ('done', 'Ready for next stage'),
         ('blocked', 'Blocked')], string='Kanban State')
+    sequence = fields.Integer(
+        string='Sequence', index=True, default=10,
+        help="Gives the sequence order when displaying a list of tickets.")
 
     def send_user_mail(self):
         self.env.ref('helpdesk_mgmt.assignment_email_template'). \
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -1,4 +1,5 @@
 from odoo import _, api, fields, models, tools
+from email.utils import getaddresses
 
 
 class HelpdeskTicket(models.Model):
@@ -7,7 +8,7 @@ class HelpdeskTicket(models.Model):
     _description = 'Helpdesk Ticket'
     _rec_name = 'number'
     _order = 'number desc'
-    _inherit = ['mail.thread', 'mail.activity.mixin']
+    _inherit = ['mail.thread', 'mail.activity.mixin', 'portal.mixin']
 
     def _get_default_stage_id(self):
         return self.env['helpdesk.ticket.stage'].search([], limit=1).id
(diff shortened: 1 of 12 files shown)


eLBati:
This can be reviewed.
Please see commits for details.
Thanks

helpdesk_mgmt/models/helpdesk_ticket_stage.py
```diff
@@ -14,6 +14,7 @@ class HelpdeskTicketStage(models.Model):
         string='Unattended')
     closed = fields.Boolean(
         string='Closed')
+    portal_user_can_close = fields.Boolean()
```
hbrunn, line 17: could you use this boolean to disallow [any logged in user to close any ticket](https://github.com/OCA/helpdesk/blob/44b501023dc4b4f06203b72be8fbfc0a16b7e64d/helpdesk_mgmt/controllers/main.py#L20)?

helpdesk_mgmt/security/ir.model.access.csv
```diff
@@ -2,7 +2,7 @@ id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_helpdesk_ticket_manager,helpdesk.ticket.manager,model_helpdesk_ticket,group_helpdesk_manager,1,1,1,1
 access_helpdesk_ticket_user,helpdesk.ticket.user,model_helpdesk_ticket,group_helpdesk_user,1,1,1,0
 access_helpdesk_ticket_user_personal,helpdesk.ticket.user.personal,model_helpdesk_ticket,group_helpdesk_user_own,1,1,1,0
-access_helpdesk_ticket_portal,helpdesk.ticket.portal,model_helpdesk_ticket,base.group_portal,1,0,0,0
+access_helpdesk_ticket_portal,helpdesk.ticket.portal,model_helpdesk_ticket,base.group_portal,1,1,0,0
```
hbrunn: this goes very far, as external users can do all kinds of mischief now by crafting json requests that to a write on tickets. Better override [`ir.attachment#check`](https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/ir_attachment.py#L338) to treat helpdesk tickets differently for portal users.

helpdesk_mgmt/security/ir.model.access.csv
```diff
@@ -3,6 +3,7 @@ access_helpdesk_ticket_manager,helpdesk.ticket.manager,model_helpdesk_ticket,gro
 access_helpdesk_ticket_user,helpdesk.ticket.user,model_helpdesk_ticket,group_helpdesk_user,1,1,1,0
 access_helpdesk_ticket_user_personal,helpdesk.ticket.user.personal,model_helpdesk_ticket,group_helpdesk_user_own,1,1,1,0
 access_helpdesk_ticket_portal,helpdesk.ticket.portal,model_helpdesk_ticket,base.group_portal,1,1,0,0
+access_helpdesk_ticket_internal_user,helpdesk.ticket.internal.user,model_helpdesk_ticket,base.group_user,1,0,0,0
```
hbrunn: better fix this one by restricting the [smart button](https://github.com/OCA/helpdesk/blob/44b501023dc4b4f06203b72be8fbfc0a16b7e64d/helpdesk_mgmt/views/res_partner_view.xml#L10) to the helpdesk user group

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -78,6 +78,8 @@ def _read_group_stage_ids(self, stages, domain, order):
         ('normal', 'Default'),
         ('done', 'Ready for next stage'),
         ('blocked', 'Blocked')], string='Kanban State')
+    sequence = fields.Integer(string='Sequence', index=True, default=10,
```
hbrunn: with the default value you basically turn off sorting by number. Better have no default, then users who use the sequence can have it, but it will be null for others and it reverts to number

hbrunn (changes requested):
it would also be good to expand on the tests in order not to decrease coverage

Also, travis [complains](https://travis-ci.com/github/OCA/helpdesk/jobs/363744110#L303)

helpdesk_mgmt/models/helpdesk_ticket.py
eLBati: If sequence is 10 for every ticket (when users don't use the sequence), tickets will be sorted by `number`, right?
(I followed the same approach as `project.task`)

helpdesk_mgmt/models/helpdesk_ticket_stage.py
eLBati: Done, thanks

helpdesk_mgmt/security/ir.model.access.csv
eLBati: Done, thanks

helpdesk_mgmt/security/ir.model.access.csv
eLBati: Restricted the smart button.
But I could not remove `access_helpdesk_ticket_internal_user` : without it, internal users keeps getting access error

eLBati:
Working on the tests

helpdesk_mgmt/models/helpdesk_ticket.py
hbrunn: good point, ignore then

helpdesk_mgmt/security/ir.model.access.csv
hbrunn: that's weird. I'll look into this and come back to you, this *should* work

eLBati:
Increased tests

helpdesk_mgmt/security/ir.model.access.csv
hbrunn: right, this only makes the button invisible. You can restrict the whole view override to the helpdesk user group in order to not have the field on the view at all.

My problem with allowing unrestricted access is that customers have the tendency to post all kinds of sensitive information in tickets. For the other groups we have `ir.rule`s to restrict this, so I could also live with this ACL if you add the internal user group to the rule restricting to own tickets.

eLBati: > You can restrict the whole view override to the helpdesk user group in order to not have the field on the view at all

Right, done, thanks

eLBati:
@hbrunn what do you think?

hbrunn:
we should find out why CI fails, many people don't review red PRs

eLBati:
@hbrunn runbot was killed. Rebuilded and everything green now

hbrunn (approved):

DarioLodeiros (approved):

PR #172, opened 2020-10-19
[FIX] Helpdesk Management Timesheet

gdgellatly:
Before this commit, in any view with ticket_id present but unused the onchange would be called and set project and task to False. It was found while trying to integrate project_timesheet_time_control with helpdesk_mgmt_timesheet, where this onchange overrides the results of default_get.

After this commit, if a helpdesk ticket is not present when calling onchange, project and task ids are not reset to False.

### DIFF
--- a/helpdesk_mgmt_timesheet/models/hr_timesheet.py
+++ b/helpdesk_mgmt_timesheet/models/hr_timesheet.py
@@ -25,5 +25,7 @@ class AccountAnalyticLine(models.Model):
     @api.onchange("ticket_id")
     def onchange_ticket_id(self):
         for record in self:
+            if not record.ticket_id:
+                continue
             record.project_id = record.ticket_id.project_id
             record.task_id = record.ticket_id.task_id


gdgellatly:
Hi @pedrobaeza @Yajo I know this isn't your module, but it is causing issues with your time control module so would really appreciate a review here.

bodedra (commented):
It's onchange. So IMO no need to multi recordsets.

How about this:

        if record.ticket_id:
            record.project_id = record.ticket_id.project_id
            record.task_id = record.ticket_id.task_id

gdgellatly:
@bodedra The code is syntactically identical. I don't think it matters but recently reviewers have shown a strong preference to reduce indentation levels even for trivial things so I did it that way. 

As an aside, there is no reason this function couldn't be called on multiple records. There is an implied api.multi so we must take care of the case.

helpdesk_mgmt_timesheet/models/hr_timesheet.py
```diff
@@ -25,5 +25,7 @@ class AccountAnalyticLine(models.Model):
     @api.onchange("ticket_id")
     def onchange_ticket_id(self):
         for record in self:
+            if not record.ticket_id:
+                continue
```
pedrobaeza, line 29:
```suggestion
        for record in self.filtered("ticket_id"):
```

gdgellatly: @pedrobaeza I consider running a filtered in an onchange an antipattern for me on small recordsets, especially for what is normally a single record. Even with non newId record it is dog slow.

```
t = env['account.analytic.line'].search([], limit=1)
>>> timeit(stmt='bool(t.ticket_id)',globals=globals())
5.048047009004222
>>> timeit(stmt='t.filtered("ticket_id")',globals=globals())
19.299567326001124
>>> timeit(stmt='for x in t:\n' '  if not x.ticket_id:\n' '    continue\n' '  pass\n',globals=globals())
6.696332884996082

>>> t = self.env['account.analytic.line'].new()
>>> timeit(stmt='for x in t:\n' '  if not x.ticket_id:\n' '    continue\n' '  pass\n',globals=globals())
7.403129160004028
>>> timeit(stmt='for x in t.filtered("ticket_id"):\n'  '  pass',globals=globals())
21.471524859000056
```

bodedra (approved):

helpdesk_mgmt_timesheet/models/hr_timesheet.py
pedrobaeza: OK, it's weird that `filtered` loses such performance...

pedrobaeza (approved):

helpdesk_mgmt_timesheet/models/hr_timesheet.py
bodedra: Curious to know, in which case `filtered` give better performance?

pedrobaeza: I expect similar performances. It's only syntactic sugar.

yajo: I guess it makes sense because with `filtered()` you must:

1. Iterate over *all* records always.
1. Produce a new recordset. I noticed it calls `browse()` without `with_prefetch`, which could mean that new queries have to be executed. I'm not sure how this works after v13 ORM refactor though.
1. Iterate again over all records in the new recordset.

I tend to avoid it these days too. Where it will be less performant the ~100% of cases is when you can do a `break` instead of a `continue`.

yajo: BTW I think https://github.com/OCA/helpdesk/pull/172#discussion_r508868961 should be reported as an issue to Odoo upstream.

PR #173, opened 2020-10-19
13.0 helpdesk time control

gdgellatly:
This is a new module, not its dependency on the prior PR, although that commit is also included here as I don't really know the process.

It is designed to complete the loop between OCA project time control and helpdesk_timesheet by autoinstalling if thos 2 modules are installed.


gdgellatly:
@pedrobaeza @Yajo This is the actual functionality I want to add for your time_control module.

pedrobaeza:
Travis is red. I talked about including directly the dependency and the functionality in the base module.

gdgellatly:
@pedrobaeza Yes it is red, I marked it draft while I work it out, however at least 1 of the test failures is due to a likely bug in how the tests are written. They compare dates after subtracting an hour, and the test ran at about 12:15 UTC. But I am kind of looking at it to determine then if the test is at fault or we have a datetime bug in the module when it crosses UTC days.

gdgellatly:
IRT to include directly in base module, I do not think it is a good idea if I am reading you correctly.  It will require a number of helpdesk modules to be installed, which is why I did it as an autoinstall if those required modules are installed.

pedrobaeza:
I mean with helpdesk_timesheet...

gdgellatly:
> I mean with helpdesk_timesheet...

Well still same argument really in the reverse, helpdesk_timesheet will need to depend on project..time_control and people may want alternative timetracking modules. I think autoinstall is probably the best option and maybe an update to the README of helpdesk_timesheet.

pedrobaeza:
OK, as you prefer. We did `crm_timesheet` directly depending on `project_timesheet_time_control` for example.

cesarlr:
@gdgellatly  @pedrobaeza  We want to backport this module to 12.0. Any comments?

pedrobaeza:
No problem in my side. Just preserve commit history.

gdgellatly:
@cesarlr No issues from me, I just saw I left a demo file in there by mistake. It has been used in production now for some time, and it works just great.

helpdesk_mgmt_timesheet_time_control/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "category": "Hidden",
+    "author": "Graeme Gellatly,Odoo Community Association (OCA)",
+    "website": "https://o4sb.com",
+    "depends": [
+        "helpdesk_mgmt_timesheet",
+        "project_timesheet_time_control",
+        "helpdesk_mgmt_project",
+    ],
+    "data": ["views/helpdesk_ticket.xml"],
+    "autoinstall": ["helpdesk_mgmt_timesheet", "project_timesheet_time_control"],
```
yajo: AFAIK it should be:

```suggestion
    "auto_install": True,
```

helpdesk_mgmt_timesheet_time_control/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2020 Graeme Gellatly
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Helpdesk Mgmt Timesheet Time Control",
+    "summary": """
+        Helpdesk Management Timesheet Time Control""",
```
yajo:
```suggestion
    "summary": "Helpdesk Management Timesheet Time Control",
```

helpdesk_mgmt_timesheet_time_control/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,39 @@
...
+        "timesheet_ids.unit_amount",
+    )
+    def _compute_show_time_control(self):
+        result = super()._compute_show_time_control()
+        for ticket in self:
+            # Never show button if timesheets are not allowed in project
+            if not (
+                ticket.project_id.allow_timesheets and ticket.team_id.allow_timesheet
+            ):
+                ticket.show_time_control = False
```
yajo:
```suggestion
            ticket.show_time_control = ticket.project_id.allow_timesheets and ticket.team_id.allow_timesheet
```

helpdesk_mgmt_timesheet_time_control/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* `O4SB <https://o4sb.com>`_:
```
yajo:
```suggestion
* `O4SB <https://o4sb.com>`__:
```

helpdesk_mgmt_timesheet_time_control/readme/INSTALL.rst
```diff
@@ -0,0 +1,8 @@
+This module auto installs when `helpdesk_mgmt_timesheet` and `project_timesheet_time_control` are
```
yajo:
```suggestion
This module auto installs when ``helpdesk_mgmt_timesheet`` and ``project_timesheet_time_control`` are
```

helpdesk_mgmt_timesheet_time_control/readme/USAGE.rst
```diff
@@ -0,0 +1,19 @@
+In general the usage instructions from `project_timesheet_time_control` apply with the following additions.
```
yajo:
```suggestion
In general the usage instructions from ``project_timesheet_time_control`` apply with the following additions.
```

helpdesk_mgmt_timesheet_time_control/views/helpdesk_ticket.xml
```diff
@@ -0,0 +1,180 @@
...
+                    string="Stop work"
+                    tabindex="-1"
+                    type="object"
+                    icon="fa-stop-circle text-warning"
+                    attrs="{'invisible': [('show_time_control', '!=', 'stop')]}"
+                    class="oe_stat_button"
+                />
+            </xpath>
+            <xpath expr="//field[@name='timesheet_ids']/tree" position="attributes">
+                <attribute name="default_order">date_time</attribute>
```
yajo, line 73:
```suggestion
                <attribute name="default_order">date_time DESC</attribute>
```

helpdesk_mgmt_timesheet_time_control/wizards/hr_timesheet_switch.py
```diff
@@ -0,0 +1,27 @@
...
+                    ("ticket_id", "=", self.env.context["active_id"]),
+                ]
+            except KeyError:
+                # If I don't know where's the user, I don't know what to suggest
+                return self.env["account.analytic.line"].browse()
+            return self.env["account.analytic.line"].search(
+                domain, order="date_time DESC", limit=1,
+            )
+        else:
+            return super()._closest_suggestion()
```
yajo:
```suggestion
    def _closest_suggestion(self):
        """Allow searching best suggestion by helpdesk.ticket."""
        result = super()._closest_suggestion()
        try:
            if not result and self.env.context["active_model"] == "helpdesk.ticket":
                return self.env["account.analytic.line"].search(
                    [
                        ("user_id", "=", self.env.user.id),
                        ("ticket_id", "=", self.env.context["active_id"]),
                    ],
                    order="date_time DESC",
                    limit=1,
                )
        except KeyError:
            # If I don't know where's the user, I don't know what to suggest
            pass
        return result
```

yajo (changes requested):
I tested it functionally and it looks very good. Some code suggestions.

gdgellatly:
@Yajo I added all the suggestions thanks. The only one I don't especially agree with was the ordering of timesheets on tickets by date_time DESC. I hate that UI on editable tree views, where the newest item is at top, but you need to add new item at bottom, but editable="top" has its own issues.  But I don't care enough to argue it, there are benefits both ways.

helpdesk_mgmt_timesheet_time_control/models/helpdesk_ticket.py
gdgellatly: @Yajo this field is not a Boolean so we can't do this.

yajo:
> The only one I don't especially agree with was the ordering of timesheets on tickets by date_time DESC. I hate that UI on editable tree views, where the newest item is at top, but you need to add new item at bottom

OK, indeed it makes more sense, you can leave that part as it was. :+1:

pedrobaeza:
For me the proper combination is editable="top" and newest on top.

yajo:
But it's true that the rest of Odoo works like @gdgellatly said AFAICR. Specifically this part of timeshee lines, at least.

pedrobaeza:
If I go to timesheets, I have the behavior I'm describing.

yajo:
No, sorry I didn't explain myself clearly. I mean when we talk about tree views that are inside form views. Such as project.task, or crm.lead (actually in crm.lead you see the weird behavior @gdgellatly said and indeed it's confusing).

pedrobaeza:
OK, now I get you. I have checked on runbot, and in tasks the order is more or less correct (most recent on bottom, and editable="bottom"), although I would prefer the reverse one, so maybe we should stick to that convention here, isn't it?

yajo:
To me, we should stick with how upstream Odoo does things. Otherwise, this produces a divergence that can stress the end user. The fact of having them on top or bottom is IMHO less important than the fact of being consistent.

I have checked how Odoo does that in tasks, in v14 upstream runbot:

imagen

They sort them top-down, and allow edits in the bottom. Then my vote goes for us doing the same.

pedrobaeza:
Yes, I agree

gdgellatly:
@pedrobaeza @Yajo Now that we all agree :) I force pushed that DESC change.

helpdesk_mgmt_timesheet_time_control/models/helpdesk_ticket.py
```diff
@@ -21,9 +21,10 @@ def _relation_with_timesheet_line(self):
     def _compute_show_time_control(self):
         result = super()._compute_show_time_control()
         for ticket in self:
-            ticket.show_time_control = (
+            if not (
                 ticket.project_id.allow_timesheets and ticket.team_id.allow_timesheet
-            )
+            ):
+                ticket.show_time_control = False
```
yajo: I guess you reverted more than you should.

yajo (changes requested):

helpdesk_mgmt_timesheet_time_control/models/helpdesk_ticket.py
gdgellatly: @Yajo nope, your suggestion to turn a selection option into a bool I didn't agree with either.

yajo: :+1: Oops I thought it *was* a bool already :open_mouth:

yajo (approved):
I suggest to squash the commits.

enricostano:
@gdgellatly @pedrobaeza should we wait for this PR to get merged before starting on the backport to v12?

cc/ @cesarlr

gdgellatly:
@enricostano This PR is basically complete and awaiting merge, just going through its rites of passage. Barring any unusual omission or oversight by me or reviewers to date, It is hugely unlikely to change.  You are probably fine to backport.

yajo:
This is just missing 1 approval to be merged.

cesarlr (approved):

enricostano:
Migrated this PR to v12 https://github.com/OCA/helpdesk/pull/201

PR #174, opened 2020-10-23
[12.0] FIX helpdesk_mgmt when creating ticket from a duplicated customer:

eLBati:
```
  File "helpdesk_mgmt/controllers/main.py", line 55, in submit_ticket
    ('email', '=', kw.get('email'))]).id
  File "odoo/odoo/fields.py", line 2900, in __get__
    raise ValueError("Expected singleton: %s" % record)
ValueError: Expected singleton: res.partner(23191, 23894)
```

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -49,10 +49,7 @@ class HelpdeskTicketController(http.Controller):
             'channel_id':
                 request.env['helpdesk.ticket.channel'].
                 sudo().search([('name', '=', 'Web')]).id,
-            'partner_id':
-                request.env['res.partner'].sudo().search([
-                    ('name', '=', kw.get('name')),
-                    ('email', '=', kw.get('email'))]).id
+            'partner_id': request.env.user.partner_id.id,
         }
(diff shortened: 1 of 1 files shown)


marcelsavegnago (approved):

dalonsod (approved):
LGTM

gdgellatly:
@eLBati are we sure this is the right way to go? I mean it resolves 1 error to potentially cause another. Given the controller is auth=user and the email field isn't editable, shouldn't request.env.user's first partner id be our first choice id? In fact it seems strange that we search on name/ allow name to be editable at all here.

gdgellatly (changes requested):
I think given that just 2 lines lower we use env.user.partner_id.ids and auth=user for the route, and email field can't be edited actually this search is somewhat pointless, and limiting to 1, potentially picks wrong partner.  For me it should just be env.user.partner_ids[0] obviously checked for empty but I think it is pretty much mandatory.

eLBati:
@gdgellatly right. Modified. Thanks

gdgellatly (approved):

Saran440 (approved):

eLBati:
merge?

PR #194, opened 2020-12-21
[IMP] add ticket number column to portal_my_tickets template

deu-dev:


dalonsod:
Interesting improvement... @DarioLodeiros travis error mentions an error related to an addon (`helpdesk_type`) that is not the modified one (`helpdesk_mgmt`). Is that correct?

PR #198, opened 2021-01-19
[13.0][FIX] Greeting in the closed notification email

teloconf:
The greeting is directed to the user instead of the customer


pedrobaeza (approved):

bodedra (approved):

pedrobaeza:
This required anyway a manual change in existing DBs, but no migration script should be provided as the template may be customized and it's worse solution.

PR #205, opened 2021-02-08
[12.0][FIX] helpdesk_mgmt_timesheet: Do not set a default project

ypapouin:
Do not set a default project if the team is not set or if timesheets are not allowed


dalonsod (approved):
LGTM. Maybe you can increase test coverage with the new cases

HaraldPanten (approved):

PR #218, opened 2021-04-15
[13.0][FIX] helpdesk_mgmt: Fix helpesk typo

MateuGForgeFlow:
@LoisRForgeFlow

### DIFF
--- a/helpdesk_mgmt/security/helpdesk_security.xml
+++ b/helpdesk_mgmt/security/helpdesk_security.xml
@@ -22,7 +22,7 @@
         </record>
     </data>
     <data noupdate="1">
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow (approved):

helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -22,7 +22,7 @@
         </record>
     </data>
     <data noupdate="1">
-        <record id="helpesk_ticket_personal_rule" model="ir.rule">
+        <record id="helpdesk_ticket_personal_rule" model="ir.rule">
```
pedrobaeza, line 25: With this change you have now duplicated the same record rule. Please don't do this thing without migration scripts.

PR #219, opened 2021-04-16
[13.0][MIG] helpdesk_mgmt: add migration script for external ID renaming

LoisRForgeFlow:
Attends https://github.com/OCA/helpdesk/pull/218#pullrequestreview-636753972.

@ForgeFlow

### DIFF
--- a/helpdesk_mgmt/migrations/13.0.1.2.3/pre-migrate.py
+++ b/helpdesk_mgmt/migrations/13.0.1.2.3/pre-migrate.py
@@ -1,0 +1,16 @@
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+xmlid_renames = [
+    (
+        "helpdesk_mgmt.helpesk_ticket_personal_rule",
+        "helpdesk_mgmt.helpdesk_ticket_personal_rule",
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_xmlids(env.cr, xmlid_renames)


helpdesk_mgmt/migrations/13.0.1.2.3/post-migrate.py
```diff
@@ -0,0 +1,46 @@
...
+    """,
+        (module, name),
+    )
+    imd_id, res_id, model = cr.fetchone()
+    cr.execute(
+        """
+        SELECT partner_id, product_id
+        FROM product_requestor_rel
+    """
+    )
```
MiquelRForgeFlow:
```suggestion
```

helpdesk_mgmt/migrations/13.0.1.2.3/post-migrate.py
```diff
@@ -0,0 +1,46 @@
...
+        DELETE FROM ir_model_data WHERE id = %s;
+    """,
+        imd_id,
+    )
+
+
+def migrate(cr, version):
+    remove_obsolete_external_id_and_record(
+        cr, "helpdesk_mgmt.helpesk_ticket_personal_rule"
+    )
```
MiquelRForgeFlow: why not use openupgradelib `delete_records_safely_by_xml_id()` method?

LoisRForgeFlow: Some people prefer not to depen on openupgrade on this "update-scripts"

MiquelRForgeFlow: You could copy the code of the method hehe

MiquelRForgeFlow (commented):
After seeing https://github.com/OCA/helpdesk/pull/218, I think the best approach is not deleting the old xmlid, but just rename the old one.

So, instead, I would run in pre-migration the `rename_xmlids` of openupgradelib. If you don't want to use openupgradelib, you can partially copy the method https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L903.

helpdesk_mgmt/migrations/13.0.1.2.3/post-migrate.py
LoisRForgeFlow: haha Copying? me? never :stuck_out_tongue: 

Just kidding, I will do that next time.

LoisRForgeFlow:
I prefer to delete it now and update version in this merge, because there is a time window (1 day now and growing until this is merged) in which someone could have already updated this module and have duplicated records.

MiquelRForgeFlow:
You can avoid the time window issue if you first check if the new one exists, then delete it. And then do the rename of the old one haha

LoisRForgeFlow:
OK

MiquelRForgeFlow (approved):
BTW, I don't want to block it. That were my suggestions :P

pedrobaeza:
If we do it quickly, we can just rename XML-IDs in pre, simplifying a lot the code, and please use openupgradelib for avoiding also this ugly code. Anyway, I'm not affected, as we don't have this module installed, but just an "external adviser".

pedrobaeza (approved):

MiquelRForgeFlow (approved):
super nice!

PR #222, opened 2021-04-20
[13.0][IMP] Improved Data Output in Team tree view

dalonsod:
Forward-port of #124 . Also fixes #220

Issue #220, opened 2021-04-16
no access to dashboard


tdombos:
The following erros message is given if a user that is the member of the User group (not Manager) tries to access the Helpdesk dashboard (app root menu): 

Sorry, you are not allowed to modify documents of type 'Helpdesk Ticket Team' (helpdesk.ticket.team). This operation is allowed for the groups: 	- Helpdesk/Helpdesk Manager - (Operation: write, User: 6)

dalonsod:
Hello @tdombos 
Which Odoo version did you try? And have you tested it on runbot, or locally? 

If it is a local environment, which Helpdesk addons are installed? And ensure that you've got the most recent versions of the addons

tdombos:
v13, I tried on runbot, and the problem is there as well.

dalonsod:
Thanks @tdombos 
Locally tested, the issue is solved making `compute_sudo=True` the `todo_ticket_count*` fields.

I've also notice that #124 is not ported to v13 yet. Those improvements could be made together with the required fix, because they are related to the same part of the code (team computed fields)

dalonsod:
Solved by #222 , this issue can be closed

PR #171, opened 2020-10-05
[FIX] Allow base user to access partner data without being in a Helpdesk group

ypapouin:
Forward port of #167

PR #229, opened 2021-05-11
[FIX][14.0] Access rights issue helpdesk_mgmt

aoso3:
Fix the problem of users with "user" access rights can't access Helpdesk dashboard.

PR:  https://github.com/OCA/helpdesk/pull/222/files

issue #220

### DIFF
--- a/helpdesk_mgmt/views/res_partner_views.xml
+++ b/helpdesk_mgmt/views/res_partner_views.xml
@@ -4,6 +4,7 @@
         <field name="name">view_partner_form</field>
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="groups_id" eval="[(4, ref('helpdesk_mgmt.group_helpdesk_user_own'))]" />
         <field name="arch" type="xml">
             <div class="oe_button_box" name="button_box">
                 <button


dalonsod:
@aoso3 could you squash commits and reword to e.g. `[FIX] Dashboard access rights issue for User role` or similar? Thanks for the contribution!

gdgellatly:
This is not the correct approach IMO, the real issue is/was writing todo_ticket_ids in the function. In any case compute_sudo is going to introduce a bug, none of the other similar models with similar access rights use it, e.g. check Projects, Journals, CRM Teams. None require compute_sudo as non stored functional fields.

dalonsod:
@aoso3 could you take a look at solution provided by @gdgellatly and also applied in #222 ?

aoso3:
@dalonsod @gdgellatly 

Sorry for the late reply.

I have removed the use of "compute_sudo" and added the missing rights to the view, can you please check the PR again.

Thank you.

gdgellatly:
@aoso3 It might pay to check, I think we fixed this ages ago

vongoethe (approved):
please merge

pedrobaeza:
It was fixed in #171, but for 13.0, so let's merge it for v14 as well:

/ocabot merge minor

marcelsavegnago (approved):

pedrobaeza:
As this has a problem with pre-commit, I have cherry-picked the v13 commit at 4f5be853b68496332e7586a2d9e61d69308fd750 and also on v15 at d9f75881d8b108d9c4839d204618bb8d3800d946

PR #231, opened 2021-05-20
[IMP][12.0] Improved all_orders_closed calculation

HekkiMelody:
Previously the method would check if the stage name was "closed" or "cancelled" but this is highly unreliable as the user can edit the names, or add more stages that are to be considered closed stages.

It is much better and more reliable to check the boolean field is_closed on the stage itself, rather than some arbitrary names.

### DIFF
--- a/helpdesk_mgmt_fieldservice/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt_fieldservice/models/helpdesk_ticket.py
@@ -78,13 +78,16 @@ class HelpdeskTicket(models.Model):
         result['views'] = [(res and res.id or False, 'form')]
         return result
 
-    @api.depends('fsm_order_ids', 'stage_id', 'fsm_order_ids.stage_id')
+    @api.depends(
+        "fsm_order_ids",
+        "fsm_order_ids.stage_id",
+        "fsm_order_ids.stage_id.is_closed",
+    )
     def _compute_all_closed(self):
         for ticket in self:
-            ticket.all_orders_closed = True
             if ticket.fsm_order_ids:
-                for order in ticket.fsm_order_ids:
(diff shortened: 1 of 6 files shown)


HekkiMelody:
Hello! Is there anyone who can tell me what I can do to improve the PR and get it accepted?

helpdesk_mgmt_fieldservice/models/helpdesk_ticket.py
```diff
@@ -84,7 +84,7 @@ def _compute_all_closed(self):
             ticket.all_orders_closed = True
             if ticket.fsm_order_ids:
                 for order in ticket.fsm_order_ids:
-                    if order.stage_id.name not in ['Closed', 'Cancelled']:
+                    if not order.stage_id.is_closed:
```
theangryangel: You could probably collapse the loop into an `any()` call, but that's more of a flavour thing (po-tay-toe, po-tah-toe)

Functionally works. I'd encourage a quick test, but given there's no coverage anyway and it's an older version of Odoo, I'm not sure what the OCA stance is on this... I'd need one of the maintainers to weigh in?

Depending on the level of brokenness, I would perhaps set up a migration to recompute the field, given the change in definition (i.e. to retrospectively fix any history tickets with incorrect values).

theangryangel (approved):

helpdesk_mgmt_fieldservice/models/helpdesk_ticket.py
HekkiMelody: I'd be more than happy to refactor the method; I remember reading somewhere that there was a strong preference for generating the smallest possible diff in a PR, so I stuck to the existing structure.

HekkiMelody:
Added a test for the method I'm rewriting

HekkiMelody:
rip this PR :D

PR #233, opened 2021-05-25
[13.0][MIG] helpdesk mgmt partner sequence

ChristianSantamaria:
Standard migration


etobella (approved):
Standard migration

raistdev (approved):

helpdesk_mgmt_partner_sequence/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,35 @@
...
+    def _prepare_ticket_number(self, values):
+        seq = False
+        if "partner_id" in values or self.partner_id:
+            partner = (
+                self.env["res.partner"].browse(values["partner_id"])
+                if "partner_id" in values
+                else self.partner_id
+            )
+            company_id = (
+                values["company_id"] if "company_id" in values else self.company_id.id
```
ValentinVinagre: wouldn't it be better?
company_id = values.get('company_id', self.company_id.id)

helpdesk_mgmt_partner_sequence/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+            # look for ticket sequence in partner
+            seq = partner.helpdesk_ticket_sequence_id
```
ValentinVinagre, line 20: in this block wouldn't it be better change the partner by commercial_partner_id?

helpdesk_mgmt_partner_sequence/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+                seq = p.helpdesk_ticket_sequence_id._get_current_sequence()
+                p.helpdesk_ticket_sequence_number_next = seq.number_next_actual
+            else:
+                p.helpdesk_ticket_sequence_number_next = None
+
+    def _inverse_seq_helpdesk_ticket_number_next(self):
+        """Inverse 'helpdesk_ticket_sequence_number_next'
+        to edit the current sequence next number.
+        """
+        for p in self:
```
ValentinVinagre: wouldn't it be better?
for p in self.filtered(lambda x: x.helpdesk_ticket_sequence_id and x.helpdesk_ticket_sequence_number_next)
and remove the if's

ValentinVinagre (commented):
Some comments.

helpdesk_mgmt_partner_sequence/models/helpdesk_ticket.py
dalonsod: I don't understand that suggested change, could you explain?

ValentinVinagre: In this case, One contact can be one sequence diferent that the parent company?

dalonsod: Yes, that's one of the functional goals:
- Use the contact ticket sequence, if defined. There's a validation if sequence company doesn't match with ticket company, in that case sequence is ignored.
- If the contact hasn't associated sequence, try the same with parent contact (company of the contact).
- Then, it's possible to use a general custom sequence for a contact company and its contacts, and for a certain contact of this company use another specific sequence.

ValentinVinagre: This functionality should be explained in the readme file.
hmm, but if I have multiple companies, this functionality doesn't work well, because I can't share the sequence with the other companies. The company_id field in the sequence is not mandatory, I would change the if to:

`if not seq.company_id or seq.company_id.id! = company_id: `

What do you think?

dalonsod: > This functionality should be explained in the readme file.

I agree

>  but if I have multiple companies, this functionality doesn't work well, because I can't share the sequence with the other companies. The company_id field in the sequence is not mandatory, I would change the if to:
>
> `if not seq.company_id or seq.company_id.id! = company_id:`
>
> What do you think?

I agree too. @ChristianSantamaria could you make the two proposed improvements?

dalonsod:
@ValentinVinagre comments attended. I've tested them on runbot, appears to work correctly. Could you check?

helpdesk_mgmt_partner_sequence/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import models
+
+
+class HelpdeskTicket(models.Model):
+    _inherit = "helpdesk.ticket"
+
+    def _prepare_ticket_number(self, values):
+        seq = False
+        if "partner_id" in values or self.partner_id:
```
ValentinVinagre: `"partner_id" in values` best with get -> values.get("partner_id")
the function 'get' is faster than 'in'

helpdesk_mgmt_partner_sequence/models/helpdesk_ticket.py
ValentinVinagre: The code it's fine ^^, You just need to add the 'functional goals' in the Readme folder.

etobella (approved):

ValentinVinagre (approved):
LGTM 👍🏻

dalonsod (approved):

AntoniRomera (approved):

AntoniRomera:
@DarioLodeiros Can you merge this?

PR #235, opened 2021-05-26
[12.0][ADD] helpdesk_mgmt_ticket_stage_auto

ChristianSantamaria:
Enables automatic stage change for tickets due to inactivity.


dalonsod:
@OCA/helpdesk-maintainers the codecov reported error is confusing for me, because it mentions `helpdesk_mgmt` addon coverage. Could you check? Thanks!

PR #239, opened 2021-06-02
[14.0][MIG]helpdesk_mgmt_timesheet

manuelregidor:
Standard migration to 14.0.


manuelregidor:
@gurneyalex Runbot has failed. I've checked the logs and I get this error:

[31mERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.

Any idea?

Thank you!

HaraldPanten (approved):
Functional review.

Could you check @ValentinVinagre @dalonsod ?

THX

dalonsod:
@HaraldPanten I'll locally test it ASAP but runbot error isn't solved yet. Who could help us with it?

HaraldPanten:
@dalonsod it seems it's only a warning. If you try the module in the runbot, it will work properly

dalonsod:
Tested on runbot LGTM, only as Admin role, because of the known bug in main addon in v13 and v14 (see #220 , #222 , and #229 ).

@manuelregidor could you add to this PR #205 , what do you think?

manuelregidor:
@dalonsod Done!

dalonsod:
@manuelregidor tested the new functionality on runbot, it seems it's not working: when a team without timesheets allowed is selected, the ticket project is unset. The same thing happens if team is unset. Could you check?

manuelregidor:
@dalonsod Could you confirm whether this is how it is supposed to work, please?
1. When a team with timesheet allowed and timesheet selected is set on a ticket, the ticket's project is set to the project selected on the team.
2. When a team with no timesheet allowed or no timesheet selected is set on a ticket, the ticket's project is unset in case a project has already been selected.
Thank you!

dalonsod:
@manuelregidor I've rebuild runbot for #205 , you can check differences between v12 and v14 current behaviors.

>     1. When a team with timesheet allowed and timesheet selected is set on a ticket, the ticket's project is set to the project selected on the team.
> 

That's OK, if the new team points to a project, use it instead of the current one.

>     2. When a team with no timesheet allowed or no timesheet selected is set on a ticket, the ticket's project is unset in case a project has already been selected.

Well, maybe you would say "When a team with no timesheet allowed or no _team_ is set on a ticket....". In those cases, the current project should remain filled, not unset. In other words, default project is only used in order to replace current ticket project if exists. And default project doesn't exist if there's no team, or team has no timesheets allowed. The goal of the PR is prevent removing project of the ticket, but still preserve replacing it for an existing one (case 1).

And there's a third case, a little bit strange: when the new team is defined with allowing timesheets... but with default project unset. In this case, project ticket is finally removed. Maybe this should be fixed in #205 , I'm not sure... @ypapouin what do you think?

manuelregidor:
@dalonsod I've compared my commit with https://github.com/OCA/helpdesk/pull/205 and I think that I might be missing something, as I don't see any differences in the way they work.

HaraldPanten:
@dalonsod Could you have a look and give us more details to solve what you suggested?

THX!

dalonsod:
@HaraldPanten too busy these days... but I'll take a look ASAP, of course!

dalonsod (approved):
Hello @manuelregidor tested last changes locally and, then, on runbot. Now works as expected: project set is only changed when a team with another default project is selected; otherwise, current project is kept.

HaraldPanten:
@pedrobaeza do you mind merging this small one, please?

THX!

pedrobaeza:
I still think this module should depend on `project_timesheet_time_control`, so I refrain from reviewing/merging it. There's also a warning on runbot that should be checked locally (logs is not available right now).

HaraldPanten:
> I still think this module should depend on `project_timesheet_time_control`, so I refrain from reviewing/merging it. There's also a warning on runbot that should be checked locally (logs is not available right now).

OK

HaraldPanten:
@DarioLodeiros @simahawk could you review/merge this one, please?

THX

simahawk:
I have no knowledge on this module nor on the reasons why @pedrobaeza would like to have it based on project_timesheet_time_control BUT:

1. here we have something working for 14 that other ppl might be interested in
2. project_timesheet_time_control is not available yet on v14 (there's a PR but it's far from being ready https://github.com/OCA/project/pull/772)

I propose to mention this option in the roadmap and merge as it is unless any blocking issue is found.

@pedrobaeza could you elaborate a bit on your proposal (if needed)?

pedrobaeza:
Such module offers the infrastructure and UX for working with timesheets, start/stop/resume timer, etc, and with very little effort. See for example https://github.com/OCA/timesheet/tree/13.0/crm_timesheet, that enables timesheet for leads. Then it's a pity to:

- Have irregular UX for one thing and the other.
- Don't have several interesting features.

HaraldPanten:
> Such module offers the infrastructure and UX for working with timesheets, start/stop/resume timer, etc, and with very little effort. See for example https://github.com/OCA/timesheet/tree/13.0/crm_timesheet, that enables timesheet for leads. Then it's a pity to:
> 
> * Have irregular UX for one thing and the other.
> * Don't have several interesting features.

Hi,

I understand your POV. Actually we only need the "basical" functionality of the module and we can not invest more time as we have migrated the other ones as well. Maybe we could open a new PR for refactoring/improving this one but it will be in a future.

IMHO these reasons are not enough for not validating/merging a migration PR, as these functionalities were not supported in previous versions.

THX!

pedrobaeza:
I'm not blocking anything, just don't participating in it. Other PSC/board member can trigger the merge.

HaraldPanten:
Ok @simahawk @pedrobaeza 

We could mention the refactoring/improvement in the ROADMAP then, before merging. As Simone suggested.

Do you agree?

AntoniRomera:
Hello,

Please note that in the 13.0 branch there are these two modules:
1. helpdesk_mgmt_timesheet_time_control 
2. helpdesk_mgmt_timesheet

It seems that this module is correctly migrated.

pedrobaeza:
For me both should be merged, but that's my opinion.

AntoniRomera:
I don't see it as a bad idea, but if so, this idea should be applied in all OCA repositories that have a timesheet and timesheet_time_control module.

pedrobaeza:
AFAIK, this is the only one that is split. `crm_timesheet` directly depends on `time_control`, and there's no more extensions for timesheets that I know.

AntoniRomera:
I will try to make another PR of the two modules together.

pedrobaeza:
Superseded by #254

PR #124, opened 2020-05-18
[IMP] Improved Data Output in Team tree view

GorkaMagra:
Improved overall data output in both the tree views of helpdesk.team, added team category tags, and the count of all the tickets assigned to that team.

PR #241, opened 2021-06-08
[FIX] 13.0 Helpdesk Management Dashboard Performance

gdgellatly:
Removes the useless related field todo_ticket_ids.

Before this commit dashboard view progressively slowed until unloadable.
After this commit - instant with 50,000 tickets.

Note the field todo_ticket_ids wasn't actually used anywhere except in computing some values and being inexplicably loaded in dashboard.

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket_team.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket_team.py
@@ -36,12 +36,7 @@ class HelpdeskTeam(models.Model):
     ticket_ids = fields.One2many(
         comodel_name="helpdesk.ticket", inverse_name="team_id", string="Tickets",
     )
-    todo_ticket_ids = fields.One2many(
-        related="ticket_ids",
-        domain="[('closed', '=', False)]",
-        string="Todo tickets",
-        readonly=True,
-    )
+
     todo_ticket_count = fields.Integer(
         string="Number of tickets", compute="_compute_todo_tickets"
     )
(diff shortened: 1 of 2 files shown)


dalonsod:
@gdgellatly could you see #124  and #222 ? Maybe it's the same fix

gdgellatly:
@dalonsod Similar but different. This is for removing a completely useless field which massively delays dashboard loading. As far as optimizing the counts, that is quite a different thing. My test on 50,000+ tickets kind of indicates that even at that level I don't have that problem so didn't bother with it.

dalonsod:
OK, I realize that the linked PRs will completely make the same as yours, adding `todo_ticket_ids` field removal. If done, the only difference should be using a recordset calculated record by record, rather than using one `read_group` function. I think the performance should be similar, but your code is more readable than `read_group` usage, I know.

Anyway, maybe you could consider adding extra existing changes in #222 that fixes #220 , is it possible?

pedrobaeza:
`read_group` always performs better, except that values are cached, which I think is not the case.

dalonsod:
> 
> 
> `read_group` always performs better, except that values are cached, which I think is not the case.

In that case, I could update #222 and remove `todo_ticket_ids` field. @gdgellatly what do you think?

gdgellatly:
> > `read_group` always performs better, except that values are cached, which I think is not the case.
> 
> In that case, I could update #222 and remove `todo_ticket_ids` field. @gdgellatly what do you think?

Fine by me.

dalonsod:
@gdgellatly done

gdgellatly:
Closing as work integrated in other PR's

PR #262, opened 2021-08-30
12.0 fix domain user

cesarlr:


cesarlr:
@nelsonramirezs @max3903 Can you review it?

helpdesk_motive/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "Helpdesk Motive",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.0.1",
```
dalonsod: @cesarlr addon version shouldn't be changed, OcaBot will do it

dalonsod (changes requested):
See comment, only code review at this moment (runbot is not still available).

helpdesk_motive/__manifest__.py
cesarlr: Ok! I've just removed the change

dalonsod (approved):
@cesarlr locally tested, LGTM.

I'm not sure if this PRs is the proper one, or if a new should be created, but maybe a good improvement IMO should be unsetting motive only if `team_id` has changed, not always. What do you think?

PR #269, opened 2021-09-16
[14.0][FIX] helpdesk_mgmt: rename already existing alias

### LINKED ISSUE #267: Installation error: the helpdesk_mgmt module conflicts with the maintenance app
I found an error during module installation. Part of the code in the file helpdesk_data.xml has conflicted with the Maintenance app:

        <record id="mail_alias_lead_info_helpdesk" model="mail.alias">
            <field name="alias_name">**helpdesk**</field>
            <field name="alias_model_id" ref="model_helpdesk_ticket" />
            <field name="alias_user_id" ref="base.user_admin" />
            <field name="alias_parent_model_id" ref="model_helpdesk_ticket_team" />
        </record>

You can change the helpdesk name to any other name to avoid conflicts (the alias_name in the model alias.mail must be unique) with the Maintenance module. Then the problem will be solved.

dalonsod:
Fixes #267

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -225,7 +225,7 @@
         </record>
         <!--default alias for leads-->
         <record id="mail_alias_lead_info_helpdesk" model="mail.alias">
-            <field name="alias_name">helpdesk</field>
+            <field name="alias_name">help</field>
             <field name="alias_model_id" ref="model_helpdesk_ticket" />
             <field name="alias_user_id" ref="base.user_admin" />
             <field name="alias_parent_model_id" ref="model_helpdesk_ticket_team" />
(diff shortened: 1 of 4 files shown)


helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -225,7 +225,7 @@
         </record>
         <!--default alias for leads-->
         <record id="mail_alias_lead_info_helpdesk" model="mail.alias">
-            <field name="alias_name">helpdesk</field>
+            <field name="alias_name">helpdesk-mgmt</field>
```
pedrobaeza: What bad decision from Odoo to use such alias, but do you think the new one is the best one? Maybe it can be simply `help`

pedrobaeza (approved):

dalonsod:
@alodoanhnhan could you review? Thanks!

dalonsod:
@manuelregidor @HaraldPanten could you review? Thanks!

HaraldPanten (approved):
Seems to be OK 👍

PR #278, opened 2021-10-18
[ADD] [14.0] helpdesk_mgmt_sale_return module

PierrickBrun:


helpdesk_mgmt_sale/models/ticket.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class HelpdeskTicket(models.Model):
+    _inherit = "helpdesk.ticket"
+
+    def _get_sale_domain(self):
+        res = "[]"
+        if not self.partner_id:
+            return res
+        return "[('partner_id', '=', self.partner_id.id)]"
```
luffah:
```python
# usually, there is no self in domain strings
# this would be cuter either with
 "[('partner_id', '=', partner_id)]"
# or with
 [('partner_id', '=', self.partner_id.id)]
 ```

File: helpdesk_mgmt_sale/models/ticket.py [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+        comodel_name="sale.order", related="ticket_id.sale_id", readonly=True
+    )
+    sale_line_id = fields.Many2one(comodel_name="sale.order.line", required=True)
+    product_id = fields.Many2one(related="sale_line_id.product_id", readonly=True)
+    product_name = fields.Char(related="product_id.name", readonly=True)
+    qty = fields.Float(string="Quantity", digits="Product Unit of Measure")
+
+    @api.onchange("sale_line_id")
+    def onchange_sale_line(self):
+        for record in self:
```
luffah, line 50: (onchange is never a multi record)
```python
def onchange(*args):
 """Return a decorator to decorate an onchange method for given fields.

   In the form views where the field appears, the method will be called
   when one of the given fields is modified. The method is invoked on a
   pseudo-record that contains the values present in the form. Field
   assignments on that record are automatically sent back to the client.

   Each argument must be a field name::

    @api.onchange('partner_id')
    def _onchange_partner(self):
        self.message = "Dear %s" % (self.partner_id.name or "")
    ...
"""
```

pedrobaeza:
I don't quite get the use case for this module. Isn't better a link between helpdesk and RMA maybe?

PierrickBrun:
> I don't quite get the use case for this module. Isn't better a link between helpdesk and RMA maybe?

Hi Pedro, the idea is to make a simple way to return a package from a helpdesk ticket. This is more designed for B2C workflows where most of the cases are the same (return several items from a sale order).

I agree that a helpdesk_rma module would be nice but in the case I want to address this would add more complexity (and a dependency)

pedrobaeza:
Well, I consider worst that you have to develop and maintain this "simple extension" (which is already 2 modules) than only linking with a consolidated system like RMA, that is maintained by my team and other contributors, and that allows your customers to evolve directly if they need something more sophisticated in any moment. If not, you'll continuously need to add small features to this set until becoming something of a similar size of RMA set.

Anyways, it's your call and OCA can admit both sets of modules. ¡And Happy Christmas!

PierrickBrun:
Hi @pedrobaeza ,

Well in my case it was easier to do it that way and I thought "Why not put it in the OCA ?" but I understand your concern. ATM there is no project to improve this further on my side.

Maybe putting a message saying that improvements shouldn't be added and that it would be better to create a helpdesk_rma module instead in the ROADMAP.rst should fix this ?


Also, my answer is very late but Happy Christmas still !

pedrobaeza:
Great, you can put that comment in the ROADMAP on the module.

PR #178, opened 2020-11-13
[FIX] helpdesk_mgmt: fix the following traceback

QaidjoharBarbhaya:
2020-11-13 13:04:21,899 23323 ERROR odoo odoo.sql_db: bad query: SELECT "res_partner".id FROM "res_partner" WHERE (("res_partner"."id" in (%s,%s))  AND  ("res_partner"."active" = %s)) ORDER BY "res_partner"."display_name"
ERROR: can't adapt type 'res.partner'
2020-11-13 13:04:21,900 23323 INFO odoo odoo.addons.fetchmail.models.fetchmail: Failed to process mail from pop server test.
Traceback (most recent call last):
  File "/home/adept-tech/odoo/odoo/addons/fetchmail/models/fetchmail.py", line 195, in fetch_mail
    res_id = MailThread.with_context(**additionnal_context).message_process(server.object_id.model, message, save_original=server.original, strip_attachments=(not server.attach))
  File "/home/adept-tech/odoo/odoo/addons/mail/models/mail_thread.py", line 1138, in message_process
    thread_id = self._message_route_process(message, msg_dict, routes)
  File "/home/adept-tech/odoo/odoo/addons/mass_mailing/models/mail_thread.py", line 28, in _message_route_process
    return super(MailThread, self)._message_route_process(message, message_dict, routes)
  File "/home/adept-tech/odoo/odoo/addons/mail/models/mail_thread.py", line 1052, in _message_route_process
    thread = ModelCtx.message_new(message_dict, custom_values)
  File "/home/adept-tech/odoo/extra-addons/helpdesk_oca/helpdesk_mgmt/models/helpdesk_ticket.py", line 242, in message_new
    ticket.message_subscribe(partner_ids)
  File "/home/adept-tech/odoo/odoo/addons/mail/models/mail_thread.py", line 2740, in message_subscribe
    partner_ids = self.env['res.partner'].sudo().search([('id', 'in', partner_ids), ('active', '=', True)]).ids
  File "/home/adept-tech/odoo/odoo/odoo/models.py", line 1643, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/home/adept-tech/odoo/odoo/odoo/addons/base/models/res_partner.py", line 710, in _search
    count=count, access_rights_uid=access_rights_uid)
  File "/home/adept-tech/odoo/odoo/odoo/models.py", line 4460, in _search
    self._cr.execute(query_str, where_clause_params)
[...]

PR #280, opened 2021-10-28
[FIX] [14.0] forward port #178 from 13.0

mtelahun:
When a new helpdesk ticket is created from an incoming email or an email reply is sent to an existing ticket the list of followers is updated. This is done by calling the method message_subscribe() with a list of partner ids. The current code tries to call this method with a list of partner objects. Because this method call happens **after** the ticket is created (for new tickets) but before it runs to completion a new identical ticket is created every time the mail is fetched from the server.

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -212,7 +212,7 @@ class HelpdeskTicket(models.Model):
             (msg.get("to") or "") + "," + (msg.get("cc") or "")
         )
         partner_ids = [
-            p
+            p.id
             for p in self.env["mail.thread"]._mail_find_partner_from_emails(
                 email_list, records=ticket, force_create=False
             )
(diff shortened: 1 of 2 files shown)


dalonsod:
Hello @metaminux , this is already resolved in v13, please take a look at #178 , maybe you should cherry pick that commit before adding the tests.

mtelahun:
Thanks, @dalonsod 
I did as you suggested, but I also kept the test.
I also changed the title to make it clear that this is a forward-port of a bug-fix in 13.0

dalonsod (approved):
👍

marcelsavegnago (approved):
LGTM

santa-asmane (approved):

PierrickBrun (approved):

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -228,7 +228,7 @@ def message_update(self, msg, update_vals=None):
             (msg.get("to") or "") + "," + (msg.get("cc") or "")
         )
         partner_ids = [
-            p
+            p.id
```
gdgellatly, line 231: Why are we even doing this?

Surely it's 

```
partner_ids = self.env["mail.thread"]._mail_find_partner_from_emails(
                email_list, records=ticket, force_create=False
            ).ids
```

Haha ignore this, I see it is the Odoo syntax. Weird why they do it that way but assume there must be a reason.

PR #281, opened 2021-11-01
14.0 [FIX] helpdesk_mgmt: onchange functions should no longer return domains

mtelahun:
In Odoo 14 onchange() methods that also return domains trigger
deprecation warnings in the logs. Instead the fields that are the
targets of the domain should use the 'domain' keyword argument
in their declarations and specify a method to call. The method
should return a standard odoo search domain string.

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -18,10 +18,18 @@ class HelpdeskTicket(models.Model):
         stage_ids = self.env["helpdesk.ticket.stage"].search([])
         return stage_ids
 
+    def _get_user_domain(self):
+        res = "[]"
+        if not self.team_id:
+            return res
+        return "[('id', 'in', self.user_ids)]"
+
     number = fields.Char(string="Ticket number", default="/", readonly=True)
     name = fields.Char(string="Title", required=True)
     description = fields.Html(required=True, sanitize_style=True)
-    user_id = fields.Many2one(comodel_name="res.users", string="Assigned user")
+    user_id = fields.Many2one(
+        comodel_name="res.users", string="Assigned user", domain=_get_user_domain
+    )
     user_ids = fields.Many2many(
         comodel_name="res.users", related="team_id.user_ids", string="Users"
     )
@@ -110,11 +118,6 @@ class HelpdeskTicket(models.Model):
     def _onchange_dominion_user_id(self):
         if self.user_id and self.user_ids and self.user_id not in self.team_id.user_ids:
             self.update({"user_id": False})
-            return {"domain": {"user_id": []}}
-        if self.team_id:
-            return {"domain": {"user_id": [("id", "in", self.user_ids.ids)]}}
-        else:
-            return {"domain": {"user_id": []}}
 
     # ---------------------------------------------------
     # CRUD
--- a/helpdesk_mgmt/tests/test_helpdesk_ticket.py
+++ b/helpdesk_mgmt/tests/test_helpdesk_ticket.py
@@ -1,6 +1,6 @@
 import time
 
-from odoo.tests import common
+from odoo.tests import common, new_test_user
 
 
 class TestHelpdeskTicket(common.SavepointCase):
@@ -8,6 +8,7 @@ class TestHelpdeskTicket(common.SavepointCase):
     def setUpClass(cls):
         super(TestHelpdeskTicket, cls).setUpClass()
         helpdesk_ticket = cls.env["helpdesk.ticket"]
+        cls.Team = cls.env["helpdesk.ticket.team"]
         cls.user_admin = cls.env.ref("base.user_root")
         cls.user_demo = cls.env.ref("base.user_demo")
         cls.stage_closed = cls.env.ref("helpdesk_mgmt.helpdesk_ticket_stage_done")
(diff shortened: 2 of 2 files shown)


PierrickBrun (approved):
LGTM

gdgellatly:
Personally I think this goes in for v15 migration. It is a very high risk/low reward PR. Anyone who has inherited this function will break. Any modules that are public will need to support both cases.

gdgellatly (changes requested):
See general comment, I just don't think the potential trouble is worth it for v14. Do it in 15.

mtelahun:
You make a good point. That hadn't occurred to me. I'm fine with putting 
it in 15.0, instead.

jacob88:
@mtelahun  @gdgellatly Yeah, looks like returning a domain in an onchange method was deprecated in Odoo v14.
https://github.com/odoo/odoo/commit/a2a1e294d7485ca839bb409129c860ad91c0b334#diff-7144f88ea32f36feb17ce1b8dda7dee1631f5ada34075414587df3948c6b3d1bR5899
https://github.com/odoo/odoo/blob/0c34fe795cee7eb6f0c5678f9e505ccc9b6e57e8/odoo/models.py#L6171-L6175

I'll give it a quick test to double check the functionality is the same using the new method, then I'll bring the commits across to the v15 migration PR.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -18,10 +18,18 @@ def _read_group_stage_ids(self, stages, domain, order):
         stage_ids = self.env["helpdesk.ticket.stage"].search([])
         return stage_ids
 
+    def _get_user_domain(self):
+        res = "[]"
+        if not self.team_id:
+            return res
+        return "[('id', 'in', self.user_ids)]"
```
jacob88, line 25: This won't work, self is not passed to the domain methods, at least not from my testing in Odoo v15.

To get this to work you'd need to do something like this.
```suggestion
    def _get_user_domain(self):
        res = []
        active_model = self.env.context.get("active_model", "helpdesk.ticket")
        active_id = self.env.context.get("active_id")
        if active_model == "helpdesk.ticket" and active_id:
            ticket = self.env[active_model].browse(active_id)
            if ticket and ticket.team_id:
                res = [('id', 'in', self.user_ids.ids)]
        return res
```
The domain is only getting picked up when the user_ids field is being read in the onchange method, this won't flow through to the frontend form view, so from a user perspective you can choose a user from the dropdown, but it's immediately deleted when the onchange fires. If you check the form view definition the domain attribute that would have been passed from the field is being set to `domain="[('share', '=', False)]"`.

Also I don't believe `self.user_ids` be evaluated within a string.

I'd suggest changing the `user_ids` field to a computed field having a fixed domain.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -18,10 +18,18 @@ def _read_group_stage_ids(self, stages, domain, order):
...
     number = fields.Char(string="Ticket number", default="/", readonly=True)
     name = fields.Char(string="Title", required=True)
     description = fields.Html(required=True, sanitize_style=True)
-    user_id = fields.Many2one(comodel_name="res.users", string="Assigned user")
+    user_id = fields.Many2one(
+        comodel_name="res.users", string="Assigned user", domain=_get_user_domain
+    )
     user_ids = fields.Many2many(
         comodel_name="res.users", related="team_id.user_ids", string="Users"
     )
```
jacob88, line 35:
```suggestion
    user_id = fields.Many2one(
        comodel_name="res.users", string="Assigned user", domain="[('share', '=', False), ('id', 'in', user_ids)]"
    )
    user_ids = fields.Many2many(
        comodel_name="res.users", string="Users", compute="_compute_user_ids"
    )
```

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -110,11 +118,6 @@ def _onchange_partner_id(self):
     def _onchange_dominion_user_id(self):
         if self.user_id and self.user_ids and self.user_id not in self.team_id.user_ids:
             self.update({"user_id": False})
```
jacob88, line 120:
```suggestion
    def _onchange_dominion_user_id(self):
        if self.user_id and self.user_ids and self.user_id not in self.user_ids._origin:
            self.update({"user_id": False})

    @api.depends('team_id', 'team_id.user_ids')
    def _compute_user_ids(self):
        all_users = self.env["res.users"].search([("share", "=", False)])
        for ticket in self:
            ticket.user_ids = ticket.team_id and ticket.team_id.user_ids or all_users
```

jacob88:
@mtelahun @gdgellatly There are some code issues to be worked out here, I think it might be a better option to drop the idea of computing the domain and compute the `user_ids` field instead.

gdgellatly:
@jacob88 that just looks hard. Firstly self must be passed, it is just whether it is effectively an object or record instance.

I'd just go something like

`domain=lambda s: "team_id and [('id', 'in',user_ids)] or []"`

I'm guessing at the syntax for lambda. I had to advise on one of these recently for a stock module in this PR. They are effectively identical problems.

https://github.com/OCA/stock-logistics-workflow/pull/894

EDIT: I had no idea we even had a user_ids field. I'm not a big fan of that. It really seems unneeded (although I haven't looked) EDIT EDIT: Of course we need it. View domains can't resolve dot syntax.

jacob88:
@gdgellatly OK, maybe if we just set the domain like this `domain="[('id', 'in', user_ids)]"` and forget about getting too complicated, the main issue here for me was that returning a domain in a onchange method had been deprecated.

Sorry for the late response, I've been rather busy the last couple of days.

PR #283, opened 2021-11-07
[15.0][MIG] helpdesk_mgmt (14.0->15.0)

jacob88:


gdgellatly:
Hi, as a minimum, if we agree that #281 is correct, then it should be part of standard migration. I'm not really familiar with the particular change to onchange domains.

jacob88:
@gdgellatly I'm happy to bring the commits across, but I am also not familiar with that particular change to base Odoo.
I'll have a look at the code and verify that a deprecation warning is raised, if it is then we might as well make the change now to minimise any breaking changes later on.

pedrobaeza:
Please cherry-pick f8ba1d9 (before the migration commit).

jacob88:
@pedrobaeza Sorry I've been quite busy the last couple of weeks.

I have brought the latest commits across from 14.0 now.

pedrobaeza (approved):

emagdalenaC2i (approved):
LGTM

PR #290, opened 2022-01-04
[13.0][FIX+IMP] helpdesk_mgmt: Use standard user assignation

pedrobaeza:
Odoo already includes a mechanism for notifying users when a document has been assigned to them, through the convention of the field name (`user_id`) and `tracking=True`.

Having to perform this manually is a lot of code to maintain, but it's also right now incorrect, as the notification is being sent also to the customer.

We switch here to the standard system.

An index is also good over this field that is widely used.

TT36502

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -9,24 +9,6 @@
     </data>
     <data noupdate="1">
         <!--Email template -->
-        <record id="assignment_email_template" model="mail.template">
-            <field name="name">Ticket Assignment</field>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Hi @sbidoul there's an error in runboat. I think it's because the certificate has expired.

victoralmau (approved):

rvalyi (approved):

chienandalu (approved):

pedrobaeza:
Cherry-pick for v14 in f8ba1d95f390d8ac570c59ba6b2bc0b06cd7b9f9

PR #292, opened 2022-01-11
[13.0][IMP] helpdesk_mgmt_rating: Only managers can change rating status

pedrobaeza:
Allowing any helpdesk user to change the rating status (to be rated or not) may lead to bad manipulations by these users, so we protect the field to only be changed by helpdesk managers.

@Tecnativa TT33737

### DIFF
--- a/helpdesk_mgmt_rating/views/helpdesk_ticket_views.xml
+++ b/helpdesk_mgmt_rating/views/helpdesk_ticket_views.xml
@@ -34,6 +34,7 @@
         <field name="arch" type="xml">
             <xpath expr="//div[hasclass('oe_title')]" position="before">
                 <field name="positive_rate_percentage" invisible="1" />
+                <field name="rating_status" invisible="1" />
             </xpath>
             <div name="button_box" position="inside">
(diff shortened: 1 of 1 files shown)


helpdesk_mgmt_rating/views/helpdesk_ticket_views.xml
```diff
@@ -34,6 +34,7 @@
         <field name="arch" type="xml">
             <xpath expr="//div[hasclass('oe_title')]" position="before">
                 <field name="positive_rate_percentage" invisible="1" />
+                <field name="rating_status" invisible="1" />
```
Tardo, line 37: Is this field really necessary in this view?

pedrobaeza: @Tardo yes, as a safeguard for possible inherited modules that use it as selector.

marcelsavegnago (approved):

chienandalu (approved):

CarlosRoca13 (approved):

Issue #296, opened 2022-02-02
[12.0] Ticket Creation e-mail beeing sent hardcoded with default template


brenoalef:
This was first reported alongside another issue in  #295 

## Module
helpdesk_mgmt

## Describe the bug
Even when there is no mail template defined for New stage, tickets created by mail send an e-mail notifying the creation to the partner.
This occurs because `helpdesk_mgmt.created_ticket_template` is being hardcoded used as template to send e-mails on ticket creation. That part of the code is not present in upper versions.

## To Reproduce
**Affected versions**:
12.0

With an Incoming Mail configuration to create Helpdesk Tickets.
1. Make sure there is no mail template defined for New stage in Configuration  
2. Send an e-mail to address for ticket creation
3. See that the partner receives an e-mail notifying the creation of the ticket

**Expected behavior**
User should only receive an e-mail if there's a template defined for the stage and, in that case, it should use the configured one.
When there's no configured template, the e-mail should not be sent.

HekkiMelody:
I'm not necessarily against this change, but it should be noted that the difference between having the email set at the stage level and this one is that as it currently exists, the email is sent on ticket creation. With your suggestion, if you were to create a ticket directly at a different stage, the email would not be sent.

Maybe a configuration setting would be better?

brenoalef:
There is currently a configuration for mail model by stage. If a ticket is created, in any stage, an e-mail is sent using the configured template. 
As it is now, even if the stage has no configured template, the "ticket created" e-mail is beeing sent. This results in two e-mails being sent if a template is defined for stage New, for example. I've tested creating a ticket at Done stage, and the result was the configured template beeing sent first and after that the hardcoded e-mail saying a ticket was created.
Creating a new configuration would be a solution, but I believe removing this part of the code in 12.0 is simpler and solves the issue.

HekkiMelody:
I don't disagree that the way the module currently behaves is problematic, I'm just pointing out that's it's not an entirely undesirable behaviour and straight-up removing it may not be the best solution.

Such use-case for one of my installations of helpdesk is that whenever a ticket is created (and it's not already on a "closed" stage on creation), a "ticket has been created" email should be sent to the customer. However, this should only happen on creation. If I were to set that "ticket has been created" email template on every single not-closed stage, I'd keep sending the "ticket has been created" email every time I advanced the ticket to a new stage, which is not what I want.

That also means that, yes, if an email is already configured for a certain stage, then two emails should be sent.

[...]

PR #298, opened 2022-02-14
[12.0][FIX] Removed hardcoded partner email on ticket creation

HekkiMelody:
Should not hardcode email sending and email template on ticket creation. The method has been removed.

I've also removed the other hardcoded method that seems to have been left over from https://github.com/OCA/helpdesk/commit/8ffd5dccda40e9534b7142099c7878b105b86761

Migration is extra safe in case the stage has been removed or another email template has been set on the stage.

Ref #296

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -110,14 +110,6 @@ class HelpdeskTicket(models.Model):
             access_rights_uid=name_get_uid)
         return self.browse(ticket_ids).name_get()
 
-    def send_user_mail(self):
-        self.env.ref('helpdesk_mgmt.assignment_email_template'). \
-            send_mail(self.id, force_send=True)
-
-    def send_partner_mail(self):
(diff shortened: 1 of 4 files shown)


helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -196,11 +188,8 @@ def create(self, vals):
 
         res = super().create(vals)
 
-        # Check if mail to the user has to be sent
-        if (vals.get('partner_id') or vals.get('partner_email')) and res:
-            res.send_partner_mail()
```
theangryangel, line 201: The only functional change I can see is that newly created "new" stages won't necessarily have any `mail_template_id` set out of the box.

It's arguable whether or not it's important to maintain that level of backwards compatibility.

Given the changes Pedro made back in https://github.com/OCA/helpdesk/commit/8ffd5dccda40e9534b7142099c7878b105b86761 (which has some overlap with this PR) I'd guess this is fine.

theangryangel (approved):

HekkiMelody:
This pr is still relevant and should not be closed.

PR #299, opened 2022-02-18
[13.0][IMP] helpdesk_mgmt: add full ticket name to customer view

mariadforgeflow:
Mails sent to customers had only the number reference of the ticket, which wasn't displayed to them in the portal view. Now both ticket number and name are always displayed in email communications and in portal.

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -46,7 +46,7 @@
                             <tbody>
                               <td valign="top" style="font-family:Arial,Helvetica,sans-serif; color: #555; font-size: 14px;">
                                 <p>Hello ${object.partner_id.name or ''},</p>
-                                <p>The ticket ${object.number} has been closed.</p>
+                                <p>The ticket "${object.display_name}" has been closed.</p>
                               </td>
                             </tbody>
                           </table>
@@ -113,7 +113,7 @@
                             <tbody>
                               <td valign="top" style="font-family:Arial,Helvetica,sans-serif; color: #555; font-size: 14px;">
                                 <p>Hello ${object.user_id.name},</p>
-                                <p>The ticket ${object.number} stage has changed to ${object.stage_id.name}.</p>
+                                <p>The ticket "${object.display_name}" stage has changed to ${object.stage_id.name}.</p>
                               </td>
                             </tbody>
                           </table>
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (approved):

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -16,7 +16,9 @@
             <field
                 name="email_cc"
             >${not object.partner_id and object.partner_email or ''|safe},</field>
-            <field name="subject">The ticket ${object.number} has been closed.</field>
+            <field
+                name="subject"
+            >The ticket "${object.display_name}" has been closed.</field>
```
pedrobaeza: I don't see that this full information should be put here.

pedrobaeza (changes requested):
Don't modify template to that long thing

helpdesk_mgmt/data/helpdesk_data.xml
mariadforgeflow: I think it is more clarifying if the name of the ticket is displayed and not only the number in the mail subject. This way, when a customer receives this notification, can know right away the ticket the mail refers to.

pedrobaeza: That's not something for the subject of the mail

LoisRForgeFlow: Ok, it is fair to not accept this change, at the end is a opinable matter so, @mariadforgeflow I think we should keep it as it is, we can change the template in our installation

pedrobaeza: Imagine a mail subject with a text + the initial subject. I think it's too much. On the body, it can be OK, but not mixed in the other sentence.

MiquelRForgeFlow:
In case of merge, use "minor" option (as the migration script has the version explicited)

MiquelRForgeFlow (approved):

helpdesk_mgmt/migrations/13.0.1.7.0/post-migrate.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_record_translations(
+        env.cr, "helpdesk_mgmt", ["closed_ticket_template", "changed_stage_template"]
```
pedrobaeza: This should go out, as it can remove legit translations (as this is executed after good translations loading). Do it on pre or not do it at all.

MiquelRForgeFlow: Ok, done in pre.

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port it to v14

PR #301, opened 2022-02-20
[14.0][FIX] Avoid error when creating ticked from a duplicated customer

JorgeQuinteros:
When the client tries to create a ticket from the portal and it has a duplicate contact:
Captura de pantalla 2022-02-20 185514

It throws the **following error**:
Captura de pantalla 2022-02-20 190445

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -54,10 +54,7 @@ class HelpdeskTicketController(http.Controller):
             .sudo()
             .search([("name", "=", "Web")])
             .id,
-            "partner_id": request.env["res.partner"]
-            .sudo()
-            .search([("name", "=", kw.get("name")), ("email", "=", kw.get("email"))])
-            .id,
+            "partner_id": request.env.user.partner_id.id,
         }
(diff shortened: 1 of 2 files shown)


Jesarregui (approved):
LGTM. Technical Review

pedrobaeza:
The submit ticket can be used without being logged in?

JorgeQuinteros:
> The submit ticket can be used without being logged in?

Hi @pedrobaeza , 
No, you must be logged in to submit a ticket.

pedrobaeza:
Then what should be removed as well is the name and email fields in such form. They should be taken from the user.

JorgeQuinteros:
> Then what should be removed as well is the name and email fields in such form. They should be taken from the user.

Yes, you are right. also the email is read only so it's a bit redundant knowing you have to log in to create a ticket. I'll update the changes tonight, thanks.

valentincastravete (approved):
LGTM! Technical Review

pedrobaeza (approved):

pedrobaeza:
Can you please fw-port it to v15?

pedrobaeza:
Cherry-picked to v13 as well at 921c9f6b926bc0752a679df488bfa99294bfa0d5

PR #302, opened 2022-02-22
[14.0][IMP] helpdesk_mgmt: add full ticket name to customer view

mariadforgeflow:
Forward port of #299 .


pedrobaeza (approved):

MiquelRForgeFlow (approved):

LoisRForgeFlow:
@mariadforgeflow The modules is already in v15, so a forward port to v15 is also needed I think.

PR #216, opened 2021-04-14
[13.0][fix] Rules on who can access to tickets from the Portal should only be governed by the ir.rules.

JordiBForgeFlow:
Fixes https://github.com/OCA/helpdesk/issues/213

@mileo @LoisRForgeFlow @pedrobaeza

PR #305, opened 2022-02-23
[14.0][FIX] helpdesk_mgmt: several forward ports

MiquelRForgeFlow:


helpdesk_mgmt/migrations/14.0.1.5.1/post-migration.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2022 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr, "helpdesk_mgmt", "migrations/14.0.1.5.1/noupdate_changes.xml"
```
pedrobaeza, line 9: In this case, I think it's better to do a replace in the string instead of losing whole template for not resetting possible customizations.

MiquelRForgeFlow: Done!

helpdesk_mgmt/controllers/myaccount.py
```diff
@@ -10,9 +10,12 @@
 class CustomerPortalHelpdesk(CustomerPortal):
     def _prepare_portal_layout_values(self):
         values = super()._prepare_portal_layout_values()
-        partner = request.env.user.partner_id
-        ticket_count = request.env["helpdesk.ticket"].search_count(
-            [("partner_id", "child_of", partner.id)]
+        ticket_count = (
```
pedrobaeza: This fix is incorrect. What has to be done is to protect the views adding the smart-button with the helpdesk group.

MiquelRForgeFlow: But you approved it in https://github.com/OCA/helpdesk/pull/216

pedrobaeza: Yeah, my bad, but it's not the right fix. Maybe I was tired of fighting everything :stuck_out_tongue_closed_eyes:

pedrobaeza: Do you understand why is not the right fix?

MiquelRForgeFlow: Fixed.

pedrobaeza:
Fast-tracking for equaling branches:

/ocabot merge patch

PR #304, opened 2022-02-22
[15.0][FIX] Avoid error when creating ticked from a duplicated customer

JorgeQuinteros:
Forward port of #301 
When the client tries to create a ticket from the portal and it has a duplicate contact:
Captura de pantalla 2022-02-20 185514

It throws the **following error**:
Captura de pantalla 2022-02-20 190445

PR #308, opened 2022-02-28
[14.0][IMP] Restore the fields in the ticket creation

JorgeQuinteros:
Restore fields that was removed in this PR: https://github.com/OCA/helpdesk/pull/304/files

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -53,6 +53,8 @@ class HelpdeskTicketController(http.Controller):
             .search([("name", "=", "Web")])
             .id,
             "partner_id": request.env.user.partner_id.id,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is there any problem with that? If they are required, maybe they shouldn't

JorgeQuinteros:
Hi @pedrobaeza ,
The way it worked until the change was that these fields were populated with the customer's information when the ticket was created.
And to preserve that functionality I think it is necessary to keep informing those fields when creating the ticket from the portal.
Because when creating the ticket from the back-end the behavior is similar, when selecting a customer, the fields are filled automatically.

pedrobaeza:
Please put such text in the commit expanded message.

JorgeQuinteros:
Done @pedrobaeza

pedrobaeza (approved):

PR #309, opened 2022-03-03
[12.0] FIX helpdesk_mgmt: 403 error when user is follower of a ticket but can't access to partner_id

eLBati:
Following the same approach as https://github.com/odoo/odoo/blob/272602193f5647f7f2270ed6ec68777625a139dd/addons/sale/controllers/portal.py#L133

### DIFF
--- a/helpdesk_mgmt/controllers/myaccount.py
+++ b/helpdesk_mgmt/controllers/myaccount.py
@@ -146,7 +146,7 @@ class CustomerPortal(CustomerPortal):
         )
         values.update({
             'date': date_begin,
-            'tickets': tickets,
+            'tickets': tickets.sudo(),
             'page_name': 'ticket',
             'pager': pager,
             'default_url': '/my/tickets',


JorgeQuinteros (approved):
LGTM! Functional & Technical Review.

Tisho99:
LGTM! Technical Review.

Tisho99 (approved):

pedrobaeza:
Steps to reproduce the problem?

eLBati:
By memory:

 - Create 2 partners, A and B
 - Create a ticket for partner A
 - Assign partner B as follower of the ticket
 - Give portal access to partner B
 - Login as partner B

Partner B can see the ticket in their ticket list but gets 403 when opening the ticket

pedrobaeza:
That should be reproducible on a test for avoiding future regressions. If not, we are adding patch over patch and not being sure the side effects of them.

JorgeQuinteros:
Hello Pedro,
It can be replicated in the runbot, in short, if the portal user is a follower of a task for which he is not the partner_id, the 403 error is thrown.
According to the registration rules, a ticket follower should be able to see a ticket that they are a follower of. https://github.com/OCA/helpdesk/commit/6dd363d7ba6205c3d77e02f9bfe687ae1a1bd7d5

Following the same logic that is done to show the Sale Order records from the portal, sudo should be added, as here:
https://github.com/odoo/odoo/blob/272602193f5647f7f2270ed6ec68777625a139dd/addons/sale/controllers/portal.py#L133

what do you think?

pedrobaeza:
Please add a test then for reproducing the problem and avoid future regressions. It can be a tour or a backend test with a portal user.

PR #312, opened 2022-04-04
[ADD] 14.0 helpdesk_mgmt_fieldservice

mtelahun:
This is a port of Open Source Integrators' module helpdesk_fieldservice for Enterprise. It allows a user to create a field-service order from a helpdesk ticket. It's been modified to function with helpdesk_mgmt in OCA. I have kept the license, which is LGPL - 3.


ivantodorovich:
Hello !
The module is already in OCA in `12.0`, so it should be a migration instead:
https://github.com/OCA/helpdesk/tree/12.0/helpdesk_mgmt_fieldservice

PR #284, opened 2021-11-11
[ADD] 14.0 helpdesk_mgmt_solution: create searchable solutions for helpdesk tickets

mtelahun:
This module is a port of #242 for 11.0 from elego Software Solutions GmbH.
It allows solutions to be attached to help desk tickets. These solutions
can then be searched from other tickets. They can be searched for by title,
content or attached tags.

PR #313, opened 2022-04-04
[14.0][MIG] helpdesk_mgmt_solution (added commit history)

MiquelRForgeFlow:
Same as https://github.com/OCA/helpdesk/pull/284 but with commit history.


elvise:
@MiquelRForgeFlow good PR! 
Do you plan to finish it ?

MiquelRForgeFlow:
@elvise Isn't it finished?

elvise:
@MiquelRForgeFlow i’m sorry i was traveling 

I would like fix the UI and UX, let me try to explain

Solution Form (view.helpdesk.solution.form):
IMG_0008

Ticket form:
There is too much space there 
IMG_0009


What do you think ?

MiquelRForgeFlow:
@elvise your suggestions have been applied :)

elvise:
> @elvise your suggestions have been applied :)

Woow thanks! :) 

If it’s possible put there summernote ;) 
IMG_0012


In the meantime  i’m testing these buttons, from user perspective looks there is something contro intuitive, i will update you soon 

IMG_0013

MiquelRForgeFlow:
> If it’s possible put there summernote ;)

I don't understand what do you mean. can you put an example?

elvise:
> > If it’s possible put there summernote ;)
> 
> I don't understand what do you mean. can you put an example?

Just the rich text editor, like this one:
IMG_0014

PR #328, opened 2022-05-30
[14.0][IMP] helpdesk_mgmt: Add partner_lang to ticket

paradoxxxzero:


hparfr (approved):

PierrickBrun (approved):

pedrobaeza (changes requested):
I don't get why don't you rely on the partner's lang

PierrickBrun:
> I don't get why don't you rely on the partner's lang

In the helpdesk the partner is not mandatory.

Either the ticket comes from a partner or it comes from someone else (for instance an in-shop buyer not registered in odoo), in that case the data about the partner (Name/email) are stored directly on the ticket.

The lang can be useful too in order to communicate with the person who created the ticket.

A partner could be created on the fly when creating a ticket but I think this would lead to a lot of duplicates and may cause problem when creating an account for someone having issued tickets in the past.

pedrobaeza:
A partner is mandatory for communicating, so that doesn't serve.

PierrickBrun:
> A partner is mandatory for communicating, so that doesn't serve.

That's right, but I think this should be done in another PR

pedrobaeza:
I would accept this if you transfer this lang when the partner is auto-created when communicating with it.

paradoxxxzero:
Unless I'm missing something it is not possible to pass anything other than an email in [the chatter suggestion list](https://github.com/odoo/odoo/blob/14.0/addons/mail/models/mail_thread.py#L1514-L1515) for a non-existing partner. As this list is used as [context for the partner creation dialog](https://github.com/odoo/odoo/blob/14.0/addons/mail/static/src/components/composer_suggested_recipient/composer_suggested_recipient.xml#L21-L28) I think we would have to patch odoo to do that.

FYI as of now the ticket `partner_name` is not used in partner creation.

Another solution would be to auto-create the partner ourselves at ticket submission but that has other implications.

hparfr:
>I would accept this if you transfer this lang when the partner is auto-created when communicating with it.

Having a ticket without a partner is a legit use case. I don't see the harm to store the lang here.

pedrobaeza:
> Having a ticket without a partner is a legit use case. I don't see the harm to store the lang here.

Yes, what is not legit is to create the partner from the ticket and to not pass such lang on the creation, and let the information de-synchronized. Please check how it is done on `crm` through default values: https://github.com/odoo/odoo/blob/9dffcc44a17fe7f5c8692c15c4463fa1bb37b60a/addons/crm/views/crm_lead_views.xml#L92-L109

hparfr:
> > Having a ticket without a partner is a legit use case. I don't see the harm to store the lang here.
> 
> Yes, what is not legit is to create the partner from the ticket and to not pass such lang on the creation, and let the information de-synchronized. Please check how it is done on `crm` through default values: https://github.com/odoo/odoo/blob/9dffcc44a17fe7f5c8692c15c4463fa1bb37b60a/addons/crm/views/crm_lead_views.xml#L92-L109

Ok. I get it now.  :-) Thanks for this precision.

PierrickBrun:
I've done the change.

Also I've removed the possibility to send an e-mail via the chatter to the partner if it is not properly created.

That way the partner is always attached to the ticket when created and has the properties passed.

If not, the partner can be created from the chatter and is only an e-mail address and is not attached to the ticket.

pedrobaeza:
But this should be done the same equal as in CRM, that handles this perfectly. Please refer to it for mimicking.

PierrickBrun:
> But this should be done the same equal as in CRM, that handles this perfectly. Please refer to it for mimicking.

This is what I've done for the partner_id field in the form.

About the chatter partner creation:

In CRM if you create a partner from the chatter the phone is not added to the partner automatically. Also, once done it is not added to the quote automatically.

See this example from odoo 14.0 runbot, in 3 steps:

Capture d’écran du 2022-09-16 15-18-21

Capture d’écran du 2022-09-16 15-18-29

Capture d’écran du 2022-09-16 15-19-10

I've added a phone number to the quote, then clicked on the partner in the chatter -> the phone number was not present.

Then, I saved and it was not added to the quote.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -243,12 +249,6 @@ def _message_get_suggested_recipients(self):
                     ticket._message_add_suggested_recipient(
                         recipients, partner=ticket.partner_id, reason=_("Customer")
                     )
-                elif ticket.partner_email:
```
pedrobaeza, line 253: Why removing this? Is it suggested by other means?

PierrickBrun: This is the code for the partner creation from the chatter, the preceding message explains why this could be problematic to keep it

pedrobaeza: I have just tried on runboat and now the mail is not suggested to be created, so that's incorrect. Please check how it's done in CRM, that the mail is suggested to be added and the partner created. You have to worry about how to transfer all the data from the helpdesk ticket form to the created partner, the same as CRM does.

PierrickBrun: Can you point me to an example of what I should copy ? I've tested on the CRM quote form and it did not do what I expected.

Indeed the partner is suggested to be added, but the partner it creates does not transfer the data from the quote and is not attached to the quote either.

pedrobaeza: Why you talk about quote? I'm talking about opportunity and sending a message from it. The recipient is suggested, and when you create the partner, all data is correctly put on the created partner.

PierrickBrun: Sorry, I was talking about the opportunity, not the quote.

I made a mistake, the partner is attached to the opportunity once the message is sent.

But the data is not put on the created partner. Here is a video of what I did (on runbot.odoo.com 14.0)

https://user-images.githubusercontent.com/8819682/190999449-f9e38806-48f3-4f05-9083-82a5482a5f5e.mp4

The data is no passed from the "phone" field. I did the test with the fields in "extra information" (Contact Name, job position & mobile) and the data is not passed either.

PierrickBrun: https://user-images.githubusercontent.com/8819682/190999856-b4e8b4ce-913d-4dd6-8289-cf28ed3a41ca.mp4

PierrickBrun: Looks like github  does not like my file, so here is a download link: https://upload.disroot.org/r/1Etx1bSy#kNF14n1+Xnr0DxwZAY5AlYaqIszYIxQi1Uax8S2a7sY=

pedrobaeza: That can be considered a bug to be notified to Odoo, but it should be the expected behavior and it should be mimicked here. You can check that it doesn't happen with the name or the email fields.

PierrickBrun: I can indeed open an issue to notify Odoo of this problem, I agree that this should be the expected behavior.

The thing is, looking at how this is done for now, this would be more of the implementation of a functionality than a bugfix. 

For now the chatter calls odoo to get the suggested recipients (https://github.com/odoo/odoo/blob/14.0/addons/mail/controllers/main.py#L314)
When Odoo answers, it passes a partner id and an email, there is no place for a list of values (https://github.com/odoo/odoo/blob/14.0/addons/mail/models/mail_thread.py#L1515)
Also, in the JS side, there is nothing other than the email that is expected, even the name is created from the email field (https://github.com/odoo/odoo/blob/14.0/addons/mail/static/src/models/thread/thread.js#L605 )

That would mean a lot of refactoring of `mail` module to make it work for the helpdesk.

It should indeed be easy to add a line in the partner creation form (https://github.com/odoo/odoo/blob/14.0/addons/mail/static/src/components/composer_suggested_recipient/composer_suggested_recipient.xml#L24 ) but as there is no way to communicate the data to put in suggestedRecipientInfo it is not feasible for now.

PierrickBrun: I've added a workaround inspired by [Opportunity to quotation](https://github.com/odoo/odoo/blob/14.0/addons/sale_crm/wizard/crm_opportunity_to_quotation.py) to send the email while making sure the partner exists

hparfr (approved):
Thanks @PierrickBrun for these improvements.
It's more aligned with other parts of odoo.

hparfr:
Hi @pedrobaeza 

Are you still requesting changes to this pr ?

pedrobaeza:
Please squash the commits into one.

paradoxxxzero:
Here you go

pedrobaeza (approved):
/ocabot merge minor

PR #330, opened 2022-06-02
[helpdesk_mgmt] Change inherit order to priorize method definitions

MTantin:
Helpdesk management is using portal mixin to ensure link between helpdesk ticket and access of them on the portal.

But i've done some tests to get access link to portal by notification email over _notify_get_groups method (recommended by Odoo).

When you override method from 'helpdesk.ticket' model, the super method called is the _notify_get_groups method from 'mail.thread' model cause this model is herited by tickets too. Finally the method with same name from portal mixin is never called!

I've updated order of inherits in 'helpdesk.ticket' model and i've the good process now:

- _notify_get_groups method call from my override code of helpdesk.ticket model
- _notify_get_groups method call from 'portal.mixin' model
- _notify_get_groups method call from 'mail.thread' model

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -8,7 +8,7 @@ class HelpdeskTicket(models.Model):
     _rec_name = "number"
     _order = "number desc"
     _mail_post_access = "read"
-    _inherit = ["mail.thread.cc", "mail.activity.mixin", "portal.mixin"]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why do you need that specific order? Steps to reproduce a problem?

MTantin:
@pedrobaeza

I've done this modification cause the _notify_get_groups method wouldn't be called without this order but i've tested again today and i don't found the same issue. I close this PR so ...

pedrobaeza:
OK, I was worried that there's a framework flaw. Let's keep it as an X file.

PR #336, opened 2022-06-20
[13.0][IMP] helpdesk_mgmt: default team specified in partner.

LoisRForgeFlow:
When creating a ticket from the portal, try to apply a default
helpdesk team based on the partner configuration.

image

Also add priority to portal view:

image

@ForgeFlow

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -42,8 +42,12 @@ class HelpdeskTicketController(http.Controller):
 
     @http.route("/submitted/ticket", type="http", auth="user", website=True, csrf=True)
     def submit_ticket(self, **kw):
+
         category = http.request.env["helpdesk.ticket.category"].browse(
-            int(kw.get("category"))
+            int(kw.get("category")))
+        team = (
+            request.env.user.partner_id.default_helpdesk_team_id
+            or request.env.user.commercial_partner_id.default_helpdesk_team_id
         )
         vals = {
             "company_id": category.company_id.id or http.request.env.user.company_id.id,
@@ -56,6 +60,8 @@ class HelpdeskTicketController(http.Controller):
             .search([("name", "=", "Web")])
             .id,
             "partner_id": request.env.user.partner_id.id,
+            "team_id": team.id,
+            "user_id": team.alias_user_id.id or team.user_id.id,
         }
         new_ticket = request.env["helpdesk.ticket"].sudo().create(vals)
         new_ticket.message_subscribe(partner_ids=request.env.user.partner_id.ids)
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -44,7 +44,8 @@ class HelpdeskTicketController(http.Controller):
     def submit_ticket(self, **kw):
 
         category = http.request.env["helpdesk.ticket.category"].browse(
-            int(kw.get("category")))
+            int(kw.get("category"))
+        )
         team = (
             request.env.user.partner_id.default_helpdesk_team_id
             or request.env.user.commercial_partner_id.default_helpdesk_team_id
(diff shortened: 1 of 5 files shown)


LoisRForgeFlow:
@pedrobaeza could you have a look?

pedrobaeza:
There are several unrelated commits. Can you split them? Also a version bump is not needed.

LoisRForgeFlow:
> There are several unrelated commits. Can you split them? Also a version bump is not needed.

Version bump removed. Regarding the split of the commits, I am not following, it is already split in 3 different commits, what else would you split?

pedrobaeza:
I mean that the PR title doesn't correspond with the contents, but I was saying about splitting in PRs, as there are some controversial changes here, so we can merge the ones that are not quickly.

LoisRForgeFlow:
> I mean that the PR title doesn't correspond with the contents, but I was saying about splitting in PRs, as there are some controversial changes here, so we can merge the ones that are not quickly.

OK, I can do that, which are the controversial ones?

pedrobaeza:
There's a lot of XML unneeded reformat, but the most controversial one is to add a new page in the already crowded partner form for just one field that is a side feature.

LoisRForgeFlow:
> There's a lot of XML unneeded reformat, but the most controversial one is to add a new page in the already crowded partner form for just one field that is a side feature.

I'm open to move it, where do you think it would fit better? In the sales & Purchases tab?

For the XML reformatting, I guess that it was pre-commit, @JasminSForgeFlow could you confirm?

JasminSForgeFlow:
> > There's a lot of XML unneeded reformat, but the most controversial one is to add a new page in the already crowded partner form for just one field that is a side feature.
> 
> I'm open to move it, where do you think it would fit better? In the sales & Purchases tab?
> 
> For the XML reformatting, I guess that it was pre-commit, @JasminSForgeFlow could you confirm?

Yes, its pre-commit

pedrobaeza:
Normal pre-commit shouldn't do that changes. Can you please restore it and split the PR as stated by pieces?

JasminSForgeFlow:
> Normal pre-commit shouldn't do that changes. Can you please restore it and split the PR as stated by pieces?

Hi,
Changes split into this new PR https://github.com/OCA/helpdesk/pull/341

Thanks

LoisRForgeFlow:
@pedrobaeza now focusing on what's remains in this PR, where should I put the helpdesk field I'm adding to the partner?

pedrobaeza:
But this seems a deformation of the way of working: teams should be redirected by alias, not by the partner that is writing to. And this will lead to contradictions: you write to an alias expecting to enter to one team (and even to have a person also assigned in the alias default values), but another team is assigned due to this isolated field.

What's the use case for that?

LoisRForgeFlow:
Hi @pedrobaeza 

The team of the alias is unchanged when receiving emails, I'm only making the field "assign to" work as intended, you would expect new tickets received via aliast to be assigned to this user (or the team leader) but they are never assigned. So it is actually a fix more than a change.

Then, users can also submit tickets via portal, in this case no team is applied and they cannot specify any in the form either. I don't think a ticket without team is very useful... so that's were this default team in the partner is used (not from the aliases). Aditionally, I apply the same logic to assign a user to the tickets coming from portal which I think it makes sense to asure that someone is notified of new tickets arriving from any source.

LoisRForgeFlow:
After discussing with @pedrobaeza he understands the point of the change but does not agree in adding a "superfluos" field. My intention with this field is to ensure that a team is assigned to tickets created from the portal (now tickets are created without any team and "lost"). This "superfluous" field would allow you to define which team to assign ticket from portal depending on who is writing to you, and therefore easily assigning it to the appropriate team that can help him/her.

One more thing to note, is that if you don't use the field, the behavior does not change at all for current instalations.

We reached a deadlock and we need more opinions.

@DarioLodeiros @eantones maybe you could give your thoughts and tell if you think this change is useful? Many thanks!

PR #338, opened 2022-06-28
[15.0][MIG] helpdesk_mgmt_timesheet : Migration to 15.0

manuelcalerosolis:
This is the migration of helpdesk_mgmt_timesheet to 15.0


dalonsod:
Hello @manuelcalerosolis thanks for the contribution!

This addon in v14 already depends on `project_timesheet_time_control` (from `OCA/project`). This addon is not migrated yet to v15, so runboat is failing, you can check it. Maybe that addon should be first migrated, I've looked at its `ROADMAP` section and a rename and movement to `OCA/timesheet` is suggested, but this task is pending too.

pedrobaeza:
The migration has been already proposed at https://github.com/OCA/project/pull/939

PR #350, opened 2022-08-04
[13.0][FIX] helpdesk_mgmt: New ticket from portal

victoralmau:
Changes done:
- Show the correct total of tickets in the portal (before, the total was filtered by those of the user's partner, but all the ones to which the user had access were shown in the list).
- Ticket creation from the portal with the category company (if set) or the user's default company.
- Redirect to the ticket created from the portal.

@Tecnativa TT38426


chienandalu (approved):
:+1: Tested. Seems to work as expected

pedrobaeza (approved):

PR #355, opened 2022-08-30
[13.0][FIX] helpdesk_mgmt: portal access error.

victoralmau:
When the user has no access rights the `search` call returns an AccessError.

@Tecnativa TT38713


CarlosRoca13 (approved):

pedrobaeza (approved):

pedrobaeza:
One thing: why using `_prepare_portal_layout_values` instead of `_prepare_home_portal_values` that is used by standard modules instead?

victoralmau:
I have applied the change to the correct function (there seems to be no reason to use the other even though they are very similar).

pedrobaeza:
I think in v12 there's no such difference. Both works, but this one avoids to gather the ticket count each time a portal page is rendered. Thanks for the changes!

/ocabot merge patch

PR #356, opened 2022-08-30
[14.0][FIX] helpdesk_mgmt: portal access error.

victoralmau:
FWP from 13.0: https://github.com/OCA/helpdesk/pull/355

When the user has no access rights the `search` call returns an AccessError.

@Tecnativa TT38713


victoralmau:
In v14 the behavior of `_prepare_home_portal_values` is different, since first the block is displayed and then the total is obtained asynchronously, this may cause the block to be displayed with total.

pedrobaeza (approved):

marcelsavegnago (approved):

PR #368, opened 2022-09-28
[12.0][FIX] Removed hardcoded partner email on ticket creation

HekkiMelody:
Reopening #298 which should not have been closed. This PR is still relevant.


pedrobaeza:
They are triaged while nobody is interested in this (specially this version is starting to be old). Anyways, check later branches, that I think has this solved, and cherry-picking that, I can fast-track it if they are equal, but reviewing new stuff for me is not possible as I'm not working anymore with this version.

HekkiMelody:
Right, I understand. Thank you for your reply @pedrobaeza

PR #371, opened 2022-10-06
[13.0][FIX] helpdesk_mgmt_crm: Prevent null value in column "notification_type"

victoralmau:
Prevent null value in column "notification_type" if message have notifications (not copied)

@Tecnativa TT38675

### DIFF
--- a/helpdesk_mgmt_crm/wizard/helpdesk_ticket_create_lead.py
+++ b/helpdesk_mgmt_crm/wizard/helpdesk_ticket_create_lead.py
@@ -39,9 +39,22 @@ class HelpdeskTicketCreateLead(models.TransientModel):
 
     def action_helpdesk_ticket_to_lead(self):
         lead = self.env["crm.lead"].create(self._prepare_vals())
+        for follower in self.ticket_id.message_follower_ids:
+            lead.message_subscribe(
+                partner_ids=[follower.partner_id.id],
+                subtype_ids=follower.subtype_ids.ids,
+            )
         self.ticket_id.write({"lead_ids": [(4, lead.id)]})
         for message in self.ticket_id.message_ids:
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
I'm not sure about this mechanism and if there's going to be again another notification to the followers. Can you be sure about it?

victoralmau:
> I'm not sure about this mechanism and if there's going to be again another notification to the followers. Can you be sure about it?

I have done some changes and improved the tests to validate as much as possible.

helpdesk_mgmt_crm/wizard/helpdesk_ticket_create_lead.py
```diff
@@ -39,9 +39,24 @@ def _prepare_vals(self):
...
+            new_message = message.copy(
+                {
+                    "model": lead._name,
+                    "res_id": lead.id,
+                    # prevent null value in column "notification_type" if message
+                    # have notifications (not copied)
+                    "notified_partner_ids": False,
+                }
+            )
+            for notification in message.notification_ids:
```
pedrobaeza: Why do you need to do this?

victoralmau: Do you think it is not necessary to transfer notifications?

pedrobaeza: I don't think so.

pedrobaeza (approved):

victoralmau:
Ping @CarlosRoca13

pedrobaeza:
Let's fast-track it for fixing the problem:

/ocabot merge patch

marcelsavegnago (commented):
LGTM

PR #341, opened 2022-07-12
[IMP] helpdesk_mgmt: Add priority in portal list and form view

JasminSForgeFlow:

PR #377, opened 2022-11-08
[IMP] helpdesk_mgmt: Modernize ticket portal views

houzefa-abba:
Bring 15.0 ticket list & form views shown in portal mode in line with those defined in the "project" module to display tasks.

While most elements remain the same, this cset also adds:
* List: group-by & search filters.
* Form: close date, assigned user, attachments.

See screenshots in https://github.com/OCA/helpdesk/pull/341#issuecomment-1240367058 for a comparison of these portal views between tickets & tasks.

Screenshots from my env, mostly demo data, hope you don't mind partial French translations:

* Ticket list, not grouped:
Screenshot at 2022-11-08 18-05-53
* Ticket list, grouped by stage:
Screenshot at 2022-11-08 18-07-28
* Ticket form, not closed, has 1 attachment:
Screenshot at 2022-11-08 18-09-38
* Ticket form, closed, multiline description with some fancy HTML, no attachments:
Screenshot at 2022-11-08 18-08-44

### DIFF
--- a/helpdesk_mgmt/controllers/myaccount.py
+++ b/helpdesk_mgmt/controllers/myaccount.py
@@ -1,13 +1,22 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
+from collections import OrderedDict
+from operator import itemgetter
+
 from odoo import _, http
 from odoo.exceptions import AccessError, MissingError
 from odoo.http import request
+from odoo.osv.expression import AND, OR
+from odoo.tools import groupby as groupbyelem
 
 from odoo.addons.portal.controllers.portal import CustomerPortal, pager as portal_pager
 
 
 class CustomerPortalHelpdesk(CustomerPortal):
+    """Routes called in portal mode to manage tickets.
+    Very similar to those in the "project" module defined to manage tasks.
+    """
+
     def _prepare_home_portal_values(self, counters):
         values = super()._prepare_home_portal_values(counters)
         if "ticket_count" in counters:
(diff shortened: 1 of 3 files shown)


helpdesk_mgmt/controllers/myaccount.py
```diff
@@ -27,70 +36,125 @@ def _prepare_home_portal_values(self, counters):
...
+        search=None,
+        search_in=None,
+        groupby=None,
+        **kw
     ):
         values = self._prepare_portal_layout_values()
-        HelpdeskTicket = request.env["helpdesk.ticket"]
-        # Avoid error if the user does not have access.
-        if not HelpdeskTicket.check_access_rights("read", raise_exception=False):
-            return request.redirect("/my")
```
victoralmau: IMO this should not be removed to avoid the error.

helpdesk_mgmt/controllers/myaccount.py
```diff
@@ -27,70 +36,125 @@ def _prepare_home_portal_values(self, counters):
...
+                ("create_date", ">", date_begin),
+                ("create_date", "<=", date_end),
+            ]
+
+        if not search_in:
+            search_in = "all"
+        if search:
+            domain += self._ticket_get_search_domain(search_in, search)
+
+        TicketSudo = request.env["helpdesk.ticket"].sudo()
```
victoralmau: I believe that you should never use `sudo()` because you will get incorrect totals for some users.

helpdesk_mgmt/controllers/myaccount.py
```diff
@@ -27,70 +36,125 @@ def _prepare_home_portal_values(self, counters):
...
+        domain = AND(
+            [
+                domain,
+                request.env["ir.rule"]._compute_domain(TicketSudo._name, "read"),
+            ]
+        )
 
         # count for pager
-        ticket_count = HelpdeskTicket.search_count(domain)
+        ticket_count = TicketSudo.search_count(domain)
```
victoralmau: I believe that you should never use sudo() because you will get incorrect totals for some users.

victoralmau (changes requested):
Thank you, visually these changes in portal are useful.

helpdesk_mgmt/controllers/myaccount.py
houzefa-abba: Thanks for the quick review

As I mentioned in the controller docstring this is inspired by the task list in the "project" module where they didn't have this; instead they filter later on with:
```python
        TicketSudo = request.env["helpdesk.ticket"].sudo()
        domain = AND(
            [
                domain,
                request.env["ir.rule"]._compute_domain(TicketSudo._name, "read"),
            ]
        )
```

This applies security rules but not ACLs. I tried disabling the ACL that gives portal users access to tickets and indeed, they can still list tickets, makes no sense. :P 

➔ I've put the redirect back & removed sudo.

I also just tried to submit a ticket with a new portal user; both portal users can't see each others' tickets now as expected.

helpdesk_mgmt/controllers/myaccount.py
houzefa-abba: Resolved with https://github.com/OCA/helpdesk/pull/377#discussion_r1017687585

helpdesk_mgmt/controllers/myaccount.py
houzefa-abba: Resolved with https://github.com/OCA/helpdesk/pull/377#discussion_r1017687585

pedrobaeza:
Please don't mix changes in the same commit and pull request. If the commit text is "Modernize ticket portal views", you can't do other changes like `sudo` or change the controller, unless it's a requirement for the views themselves.

houzefa-abba:
> Please don't mix changes in the same commit and pull request. If the commit text is "Modernize ticket portal views", you can't do other changes like `sudo` or change the controller, unless it's a requirement for the views themselves.

Controller changes here are needed to add search input & group-by which the modern view relies on.
You are right about the sudo change, it was mixed in the same place and I didn't think much about it; good thing it got caught.

houzefa-abba:
I noticed code coverage alerts in the CI - I'm thinking about adding portal tests similar to those in Odoo's project module, just not sure I'll be able to today

pedrobaeza:
That would be great, thanks.

houzefa-abba:
Rebased on top of https://github.com/OCA/helpdesk/pull/378

houzefa-abba:
In my previous rebase on https://github.com/OCA/helpdesk/pull/378 coverage tests were green but now I see they are red, I guess we were flirting with limits

I'm thinking of adding a commit to cover new filters / group-by code added in the controller

pedrobaeza (approved):

pedrobaeza:
No worry about codecov, as although interesting, it's not mandatory:

/ocabot merge major

PR #378, opened 2022-11-09
[IMP] helpdesk_mgmt: Test portal routes

houzefa-abba:
This is a start to increase coverage; not exhaustive yet.

Note the "project" module & some others in Odoo now use a tour spec defined in a js file instead to test their controllers; I didn't go that far here.

Test list, form, new ticket, unauthorized redirect.

### DIFF
--- a/helpdesk_mgmt/tests/test_portal.py
+++ b/helpdesk_mgmt/tests/test_portal.py
@@ -1,0 +1,136 @@
+from odoo import http
+from odoo.tests.common import new_test_user
+
+from odoo.addons.base.tests.common import HttpCaseWithUserPortal
+
+
+class TestPortal(HttpCaseWithUserPortal):
+    """Test controllers defined for portal mode.
+    This is mostly for basic coverage; we don't go as far as fully validating
+    HTML produced by our routes.
+    """
+
+    def setUp(self):
+        super().setUp()
+        # Create a basic user with no helpdesk permissions.
+        new_test_user(self.env, login="test-basic-user")
+        # Create a ticket submitted by our portal user.
+        self.portal_ticket = self._create_ticket(
+            self.partner_portal, "portal-ticket-title"
+        )
+
+    def test_ticket_list(self):
+        """List tickets in portal mode, ensure it contains our test ticket."""
+        self.authenticate("portal", "portal")
+        resp = self.url_open("/my/tickets")
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn("portal-ticket-title", resp.text)
+
+    def test_ticket_form(self):
+        """Open our test ticket in portal mode."""
+        self.authenticate("portal", "portal")
+        resp = self.url_open(f"/my/ticket/{self.portal_ticket.id}")
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn("portal-ticket-title", resp.text)
+        self.assertIn('<h4 class="page-header">History</h4>', resp.text)
+
+    def test_submit_ticket(self):
+        """Submit a ticket in portal mode."""
+        new_ticket_title = "portal-new-submitted-ticket-subject"
+
+        def count_new_tickets():
+            return self.env["helpdesk.ticket"].search_count(
+                [("name", "=", new_ticket_title)]
+            )
+
+        self.assertEqual(count_new_tickets(), 0)  # not found so far
+
+        self.authenticate("portal", "portal")
+        resp = self.url_open(
+            "/submitted/ticket",
+            data={
+                "category": self.env.ref("helpdesk_mgmt.helpdesk_category_1").id,
+                "csrf_token": http.WebRequest.csrf_token(self),
+                "subject": new_ticket_title,
+                "description": "portal-new-submitted-ticket-description",
+            },
+        )
+        self.assertEqual(resp.status_code, 200)
+        self.assertEqual(count_new_tickets(), 1)  # new ticket found
+
+    def test_ticket_list_unauthenticated(self):
+        """Attempt to list tickets without auth, ensure we get sent back to login."""
+        resp = self.url_open("/my/tickets", allow_redirects=False)
+        self.assertEqual(resp.status_code, 303)
+        self.assertTrue(resp.is_redirect)
+        # http://127.0.0.1:8069/web/login?redirect=http%3A%2F%2F127.0.0.1%3A8069%2Fmy%2Ftickets
+        self.assertIn("/web/login", resp.headers["Location"])
+
+    def test_ticket_list_unauthorized(self):
+        """Attempt to list tickets without helpdesk permissions, ensure we get
+        sent back to /my.
+        """
+        self.authenticate("test-basic-user", "test-basic-user")
+        resp = self.url_open("/my/tickets", allow_redirects=False)
+        self.assertEqual(resp.status_code, 303)
+        self.assertTrue(resp.is_redirect)
+        # http://127.0.0.1:8069/my
+        self.assertTrue(resp.headers["Location"].endswith("/my"))
+
+    def test_tickets_2_users(self):
+        """Check tickets between 2 portal users; ensure they can't access each
+        others' tickets.
+        """
+        portal_user_1 = self.user_portal  # created by HttpCaseWithUserPortal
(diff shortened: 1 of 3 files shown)


helpdesk_mgmt/tests/test_portal.py
```diff
@@ -0,0 +1,76 @@
...
+class TestPortal(HttpCaseWithUserPortal):
+    """Test controllers defined for portal mode.
+    This is mostly for basic coverage; we don't go as far as fully validating
+    HTML produced by our routes.
+    """
+
+    def setUp(self):
+        """Override test set-up to prepare data used in tests."""
+
+        super(TestPortal, self).setUp()
```
vincent-hatakeyama:
```suggestion
        super().setUp()
```

houzefa-abba:
Rereading this, I note my `test_ticket_list_unauthorized` is not checking authorization but it is only checking authentication instead - I'll improve & also add a check to ensure one user can't read a ticket sent by another user

houzefa-abba:
> Rereading this, I note my `test_ticket_list_unauthorized` is not checking authorization but it is only checking authentication instead - I'll improve & also add a check to ensure one user can't read a ticket sent by another user

This is done in the version I just pushed, see `test_ticket_list_unauthenticated` / `test_ticket_list_unauthorized` / `test_tickets_2_users`

[codecov details](https://app.codecov.io/gh/OCA/helpdesk/commit/64e8a886596924f95308b40a5356eca3f398c815) aren't loading right now but I'm expecting `controllers/myaccount.py > Avoid error if the user does not have access.` will now be covered.

I'll be back on Monday, please have a look @OCA/helpdesk-maintainers :)

helpdesk_mgmt/tests/test_portal.py
```diff
@@ -0,0 +1,124 @@
...
+    """Test controllers defined for portal mode.
+    This is mostly for basic coverage; we don't go as far as fully validating
+    HTML produced by our routes.
+    """
+
+    def setUp(self):
+        """Override test set-up to prepare data used in tests."""
+
+        super().setUp()
+
```
victoralmau:
```suggestion
        new_test_user(
            self.env,
            login="test-basic-user",
        )

```

helpdesk_mgmt/tests/test_portal.py
```diff
@@ -0,0 +1,124 @@
...
+        # http://127.0.0.1:8069/web/login?redirect=http%3A%2F%2F127.0.0.1%3A8069%2Fmy%2Ftickets
+        self.assertIn("/web/login", resp.headers["Location"])
+
+    def test_ticket_list_unauthorized(self):
+        """Attempt to list tickets without ticket ACL, ensure we get sent back
+        to /my.
+        """
+
+        self.env.ref("helpdesk_mgmt.access_helpdesk_ticket_portal").perm_read = False
+        self.authenticate("portal", "portal")
```
victoralmau:
```suggestion
        self.authenticate("test-basic-user", "test-basic-user")
```

victoralmau: I think it is better to test it with a user without permissions in helpdesk

helpdesk_mgmt/tests/test_portal.py
```diff
@@ -0,0 +1,124 @@
...
+
+        self._create_ticket(portal_user_1.partner_id, "ticket-user-1")
+        self._create_ticket(portal_user_2.partner_id, "ticket-user-2")
+
+        # portal_user_1 only sees ticket_1
+        self.authenticate("portal", "portal")
+        resp = self.url_open("/my/tickets")
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn("ticket-user-1", resp.text)
+        self.assertNotIn("ticket-user-2", resp.text)
```
victoralmau, line 97:
```suggestion
        self.assertNotIn("ticket-user-2", resp.text)
        resp = self.url_open(f"/my/ticket/{ticket_1.id}")
        self.assertEqual(resp.status_code, 200)
        self.assertIn("ticket-user-1", resp.text)
        resp = self.url_open(f"/my/ticket/{ticket_2.id}")
        self.assertEqual(resp.status_code, 301)
```
And something similar to portal2 user.

helpdesk_mgmt/tests/test_portal.py
```diff
@@ -0,0 +1,124 @@
...
+        others' tickets.
+        """
+
+        portal_user_1 = self.user_portal  # created by HttpCaseWithUserPortal
+        portal_user_2 = self.user_portal.copy(
+            default={"login": "portal2", "password": "portal2"}
+        )
+
+        self._create_ticket(portal_user_1.partner_id, "ticket-user-1")
+        self._create_ticket(portal_user_2.partner_id, "ticket-user-2")
```
victoralmau:
```suggestion
        ticket_1 = self._create_ticket(portal_user_1.partner_id, "ticket-user-1")
        ticket_2 = self._create_ticket(portal_user_2.partner_id, "ticket-user-2")
```

victoralmau (changes requested):

helpdesk_mgmt/tests/test_portal.py
houzefa-abba: Ok yes I was testing access rule via ticket list but not ticket form, will add

helpdesk_mgmt/tests/test_portal.py
houzefa-abba: In portal mode, one can't customize security groups; our "portal" demo user only has the portal security group, hence this ACL change

~~That test-basic-user would not have portal access at all so it is not testing the same thing ➔ I'll add but as a separate test~~
Scratch that, I can see a new basic user still has a portal, just no tickets displayed in there ➔ I'll replace my test the way you suggest

houzefa-abba:
Done @victoralmau 

I also changed ADD to IMP in the commit msg as [according to odoo doc](https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html#tag-and-module-name) this cset is not adding a new module

victoralmau (approved):
Code review OK.

helpdesk_mgmt/tests/test_portal.py
```diff
@@ -0,0 +1,147 @@
...
+
+class TestPortal(HttpCaseWithUserPortal):
+    """Test controllers defined for portal mode.
+    This is mostly for basic coverage; we don't go as far as fully validating
+    HTML produced by our routes.
+    """
+
+    def setUp(self):
+        """Override test set-up to prepare data used in tests."""
+
```
pedrobaeza: Please avoid empty lines inside the method. They are discouraged to be used.

houzefa-abba: Ok I can see in partner_firstname tests we don't add new lines after docstrings ➔ I have removed all NLs after docstrings

Also removed that setUp docstring as it is too usual to need explanation, I can see other tests never document it

The only NLs remaining are those that make sense inside longer tests (such as `test_tickets_2_users`)

I didn't see this mentioned on https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst would be interesting to add I think :)

pedrobaeza: Yeah, docstrings are very useful, specially if the test name is not descriptive enough. About including this in the guidelines, the idea is to avoid to repeat all the Python conventions. Specifically, this one is already pointed in the old PEP8 one: https://peps.python.org/pep-0008/#blank-lines

This was later a bit nuanced, as having empty lines inside the method, even if logical groups, is worse than having a comment line saying what the logical group does, so that's also my preferred option.

pedrobaeza:
The idea behind of using tour tests (with JS), is to replicate the navigation behavior of a user, so they are very advisable, but well, these ones are a first step. Please add the point in the ROADMAP for adding the other tour tests.

pedrobaeza (approved):

houzefa-abba:
> replicate the navigation behavior of a user

Yes good I wanted a place to add that TODO besides this PR ➔ added

I didn't know which tag to use for that commit, I used "UPD" which the bot uses when updating other READMEs, I hope that's fine

pedrobaeza:
Thanks for this and keep up the good work, Houzéfa!

PR #379, opened 2022-11-15
[15.0][FIX] helpdesk_mgmt: Preserve line breaks

houzefa-abba:
on tickets submitted from the portal.

Rely on `plaintext2html`, same as Odoo in its controller for message threads (`/odoo15/addons/portal/controllers/mail.py`).

Existing portal test adapted to check this.


bodedra (approved):
LG 👍

pedrobaeza (approved):

PR #394, opened 2022-12-21
[13.0][IMP] helpdesk_mgmt: Add team to new ticket form portal page

victoralmau:
Add team field to new ticket form portal page.

@Tecnativa TT40877


helpdesk_mgmt/controllers/main.py
```diff
@@ -28,6 +28,13 @@ def support_ticket_close(self, **kw):
 
         return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
 
+    def _get_teams(self):
+        return (
+            http.request.env["helpdesk.ticket.team"].search([("active", "=", True)])
```
chienandalu: Probably you need to call this with `sudo`:

image

helpdesk_mgmt/controllers/main.py
```diff
@@ -57,6 +70,9 @@ def submit_ticket(self, **kw):
             .id,
             "partner_id": request.env.user.partner_id.id,
         }
+        if company.helpdesk_mgmt_new_ticket_teams:
+            team = http.request.env["helpdesk.ticket.team"].browse(int(kw.get("team")))
```
chienandalu: This lets the user to set whatever team he please. Maybe we should check that it's a valid team.

chienandalu (changes requested):
Some comments

helpdesk_mgmt/controllers/main.py
victoralmau: I understand what you mean, but I don't think it's necessary.
Let's say I want to create a ticket to the support team (to which I don't belong), I should be able to create it, don't you think?

chienandalu: It isn't critical indeed but it's a good custom to be picky with user input :sweat_smile: In this case a user could create a ticket in an archived team (randomly picking the team id or knowing it before hand).

chienandalu (approved):
Tested :slightly_smiling_face: :+1:

pedrobaeza:
Another needed feature is a check at team level to be shown or not on the portal.

victoralmau:
> Another needed feature is a check at team level to be shown or not on the portal.

Done.

helpdesk_mgmt/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    helpdesk_mgmt_new_ticket_teams = fields.Boolean(string="Teams")
```
pedrobaeza:
```suggestion
    helpdesk_mgmt_portal_select_team = fields.Boolean(string="Select team in Helpdesk portal")
```

helpdesk_mgmt/models/res_config_settings.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    helpdesk_mgmt_new_ticket_teams = fields.Boolean(
```
pedrobaeza: The same name change.

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -49,6 +49,7 @@ class HelpdeskTeam(models.Model):
     todo_ticket_count_high_priority = fields.Integer(
         string="Number of tickets in high priority", compute="_compute_todo_tickets"
     )
+    show_in_portal = fields.Boolean(string="Show in portal form", default=True)
```
pedrobaeza:
```suggestion
    show_in_portal = fields.Boolean(string="Show in portal form", default=True, help="Allow to select this team when creating a new ticket in the portal.")
```

victoralmau:
Changes done.

helpdesk_mgmt/controllers/main.py
pedrobaeza: What about this?

```suggestion
            team = http.request.env["helpdesk.ticket.team"].search([("id", "=", int(kw.get("team")), ("show_in_portal", "=", True)])
```

pedrobaeza (approved):

PR #395, opened 2022-12-22
[13.0][IMP] helpdesk_mgmt: User: Team tickets group

victoralmau:
User: Team tickets group

@Tecnativa TT40878


chienandalu (changes requested):
Can you update the readme with this new setting?

chienandalu (approved):

helpdesk_mgmt/README.rst
```diff
@@ -115,6 +115,8 @@ Tags
 Usage
 =====
 
+#. Different user permissions are available to restrict access to tickets: User: Personal tickets, User: Team tickets and User.
```
pedrobaeza: This is configuration, not usage, and please explain the goal of each permission.

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -9,7 +9,13 @@ class HelpdeskTeam(models.Model):
     _inherit = ["mail.thread", "mail.alias.mixin"]
 
     name = fields.Char(string="Name", required=True)
-    user_ids = fields.Many2many(comodel_name="res.users", string="Members")
+    user_ids = fields.One2many(
```
pedrobaeza: You can belong to more than one team, so this is not correct.

pedrobaeza (changes requested):

helpdesk_mgmt/README.rst
victoralmau: IMO it is not necessary to add this because I think the name of the groups is obvious, what do you think?

pedrobaeza: Well, it's not bad to explain a bit more apart from the name. Sometimes the obvious for us is not the same for the plain user.

helpdesk_mgmt/readme/CONFIGURE.rst
```diff
@@ -74,3 +74,13 @@ Tags
...
    :alt: Ticket Tags
    :width: 600 px
+
+
+Permissions
+~~~~~~~~~~~
+
+There are restrictions to read tickets according to the user's permissions set in Helpdesk.
+
+#. *User: Personal tickets*: User can list his tickets (those that are assigned to his user) or those that are not assigned to anyone.
```
pedrobaeza: Use gender neutral words:

```suggestion
#. *User: Personal tickets*: User can list their tickets (those that are assigned to their user) or those that are not assigned to anyone.
```

pedrobaeza: And this should be that the tickets are only visible if no team nor user is assigned.

helpdesk_mgmt/readme/CONFIGURE.rst
```diff
@@ -74,3 +74,13 @@ Tags
...
+#. *User: Team tickets*: User can list the tickets that are assigned to the teams to which he/she belongs or the tickets that are not assigned to any team.
```
pedrobaeza: It should be that are not assigned to any team nor user.

helpdesk_mgmt/readme/CONFIGURE.rst
```diff
@@ -74,3 +74,13 @@ Tags
...
+#. *User*: User can list all tickets.
```
pedrobaeza: Change better to `is able to see` on all the lines.

```suggestion
#. *User*: User is able to see all the tickets.
```

helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -6,10 +6,15 @@
...
+        <record id="group_helpdesk_user_team" model="res.groups">
+            <field name="name">User: Team tickets</field>
+            <field name="category_id" ref="module_helpdesk_category" />
+            <field name="implied_ids" eval="[(4, ref('group_helpdesk_user_own'))]" />
+        </record>
         <record id="group_helpdesk_user" model="res.groups">
             <field name="name">User</field>
             <field name="category_id" ref="module_helpdesk_category" />
-            <field name="implied_ids" eval="[(4, ref('group_helpdesk_user_own'))]" />
+            <field name="implied_ids" eval="[(4, ref('group_helpdesk_user_team'))]" />
```
pedrobaeza, line 17: This change requires a migration script.

helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -154,7 +154,7 @@
                             <field name="user_ids" invisible="1" readonly="1" />
                             <field
                                 name="user_id"
-                                domain="[('share', '=', False)]"
+                                domain="[('share', '=', False),('id', 'in', user_ids)]"
```
pedrobaeza: Please put at field level instead, and it should be:
```
domain="team_id and [('share', '=', False),('id', 'in', user_ids)] or [('share', '=', False)]"
```

helpdesk_mgmt/models/res_users.py
```diff
@@ -0,0 +1,12 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    helpdesk_team_ids = fields.Many2many(
```
pedrobaeza, line 8: Why is this needed?

pedrobaeza (changes requested):
You should handle also if a user is assigned and you change the team, that the user should be kept only if it belongs to that team.

helpdesk_mgmt/models/res_users.py
victoralmau: This is necessary for the rule (`ir.rule`). Another possibility is to use `user_ids` field of tickets, which option do you think is better?

pedrobaeza: We should balance performance and complexity. The question here is that this field is not computed, so how it's filled?

victoralmau: This field comes from https://github.com/OCA/helpdesk/blob/13.0/helpdesk_mgmt/models/helpdesk_ticket_team.py#L12

pedrobaeza: OK, please put a comment about that, and fix in both the name of the columns and table for avoiding surprises in the future.

victoralmau:
All suggested changes done.

pedrobaeza (approved):
/ocabot merge major

PR #399, opened 2022-12-22
[14.0][ADD] Helpdesk Stage Validation

BojanOD:
Help Desk stage validation module that checks if certain fields are populated with data before switching to the desired stage from the model `helpdesk.ticket.stage`.


BojanOD:
Hi @pedrobaeza @ivantodorovich can you review this PR, thank you

damdam-s (approved):

helpdesk_mgmt_stage_validation/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,14 @@
...
+
+from .validate_utils import validate_stage_fields
+
+
+class HelpdeskTicket(models.Model):
+    _inherit = "helpdesk.ticket"
+
+    @api.constrains("stage_id")
+    def _validate_stage_fields(self):
+        validate_stage_fields(self)
```
jcoux: Here is the only location where you use the function `validate_stage_fields`.
For me it's better to have directly here the function: it's more easy to read/maintain, and extract the function have no interest.

helpdesk_mgmt_stage_validation/models/validate_utils.py
```diff
@@ -0,0 +1,21 @@
...
+        field_ids = stage.validate_field_ids
+        field_names = [x.name for x in field_ids]
+        values = rec.read(field_names)
+        for name in field_names:
+            if not values[0][name]:
+                raise ValidationError(
+                    _(
+                        'Cannot move to stage "%s" '
+                        'until the "%s" field is set.' % (stage.name, name)
+                    )
```
jcoux: Don't do string replacement into translate function, even if the translations can't be generated correctly.

```suggestion
                    _(
                        'Cannot move to stage "%s" '
                        'until the "%s" field is set.'
                    )  % (stage.name, name)
```

And please add the ticket name into the error message.

And if you can give all the missing fields on all tickets in 1 message it can be cool.
Even if, if you have by example 5 missing fields on 3 tickets, you will have 15 exceptions one by one after fixing each one by one.

helpdesk_mgmt_stage_validation/tests/test_helpdesk_ticket_stage_validation.py
```diff
@@ -0,0 +1,51 @@
...
+        field_name = fields.first(stage.validate_field_ids).name
+        return 'Cannot move to stage "%s" until the "%s" field is set.' % (
+            stage_name,
+            field_name,
+        )
+
+    def test_helpdesk_ticket_stage_validation(self):
+        validate_message = self.get_validate_message(self.stage_ticket_assigned)
+        with self.assertRaisesRegex(ValidationError, validate_message):
+            self.ticket.write({"stage_id": self.stage_ticket_assigned.id})
```
jcoux, line 61: Please also test that with setting `assigned_date`, we don't have exception (to be sure the normal process still work).

jcoux (changes requested):

BojanOD:
@jcoux I covered all of the requested changes.

helpdesk_mgmt_stage_validation/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,47 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class HelpdeskTicket(models.Model):
+    _inherit = "helpdesk.ticket"
+
+    def _check_ticket_has_empty_fields(self, stage):
```
jcoux: No need to pass stage as argument, you can have it with `self.stage_id`.

```suggestion
    def _check_ticket_has_empty_fields(self):
```

helpdesk_mgmt_stage_validation/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,47 @@
...
+        self.ensure_one()
+        error_message = False
+        stage = stage
```
jcoux:
```suggestion
```

helpdesk_mgmt_stage_validation/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,47 @@
...
+        field_ids = stage.validate_field_ids
+        field_names = [x.name for x in field_ids]
+        values = self.read(field_names)
+        for name in field_names:
+            if not values[0][name]:
+                fields = [
+                    field_name
+                    for field_name in field_names
+                    if not values[0][field_name]
```
jcoux: Extract the `fields` computation of the loop, and you can remove the loop.
```
fields = [...]
if fields:
    error_message = _(...)
    return error_message
```

helpdesk_mgmt_stage_validation/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,47 @@
...
+                ]
+                fields = ", ".join(fields)
+                error_message = _(
+                    "Ticket %s can't be moved to the stage %s until "
+                    "the following fields are set: %s."
+                    % (self.name, stage.name, fields)
+                )
```
jcoux:
```suggestion
                error_message = _(
                    "Ticket %s can't be moved to the stage %s until "
                    "the following fields are set: %s."
                ) % (self.name, stage.name, fields)
```

jcoux (changes requested):

jcoux (approved):

elvise:
@BojanOD good idea!

For more flexibility can you use the “widget domain” ?

elvise:
@BojanOD can you rebase for testing purpose ?

elvise:
@BojanOD can you rebase ?

PR #403, opened 2023-01-02
[14.0][ADD] Helpdesk Stage Server Action

BojanOD:
A module that allows a user to set a server action for a `helpdesk.ticket.stage`.

Once a ticket is switched to a different stage, if that stage has a server action assigned to it, then that server action will be executed.


damdam-s (approved):

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,37 @@
...
+        return records
+
+    def write(self, vals):
+        res = super().write(vals)
+        action = (
+            self.env["helpdesk.ticket.stage"].browse(vals["stage_id"]).action_id
+            if "stage_id" in vals
+            else None
+        )
+        if action:
```
jcoux: We don't want that, because here, you execute the action each time we have `stage_id` in `vals`.
But if you call write function with the current `stage_id`, you will execute the action even if finally, the stage didn't changed.
Then please execute the action only if the stage really change.

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket_stage.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HelpdeskTicketStage(models.Model):
+    _inherit = "helpdesk.ticket.stage"
+
+    action_id = fields.Many2one("ir.actions.server", string="Server Action")
```
jcoux: Please add a help section to explain how the field is used, and which values in passed in context of the action (on model `helpdesk.ticket` and with ticket ids).

Also I would have a restriction here to have only actions with `model_id` = `helpdesk.ticket`.

helpdesk_mgmt_stage_server_action/tests/test_helpdesk_ticket_stage.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+        cls.helpdesk_ticket_stage_2 = cls.HelpdeskTicketStage.create(
+            {
+                "name": "Stage 2",
+                "action_id": cls.server_action_helpdesk_ticket.id,
+                "sequence": 2,
+            }
+        )
+
+    def test_helpdesk_ticket(self):
```
jcoux: Please complete unit tests with different cases we can have (list non exhaustive):
* action on stage when creating the ticket
* action with stage when write ticket, but with current stage (no real stage change)
* ...

And please also add test without action to check no action has been executed.

helpdesk_mgmt_stage_server_action/tests/test_helpdesk_ticket_stage.py
```diff
@@ -0,0 +1,53 @@
...
+        self.helpdesk_ticket = self.HelpdeskTicket.create(
+            {
+                "name": "Helpdesk Ticket",
+                "stage_id": self.helpdesk_ticket_stage_1.id,
+                "description": "Helpdesk Ticket Description",
+            }
+        )
+        self.helpdesk_ticket.write({"stage_id": self.helpdesk_ticket_stage_2.id})
+        self.assertEqual(self.helpdesk_ticket.user_id, self.env.user)
```
jcoux: Please also test that before write, the user is not already equal to `self.env.user`.
To be sure the user has really changed.

jcoux (changes requested):

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,37 @@
...
+            if action:
+                context = {
+                    "active_model": self._name,
+                    "active_ids": [record.id],
+                }
+                action.with_context(**context).run()
+        return records
+
+    def write(self, vals):
+        records = self.filtered(lambda l: l.stage_id.id != vals.get("stage_id"))
```
jcoux:
```suggestion
        records = "stage_id" in vals and self.filtered(lambda l: l.stage_id.id != vals.get("stage_id"))
```

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,37 @@
...
+        res = super().write(vals)
+        action = (
+            self.env["helpdesk.ticket.stage"].browse(vals["stage_id"]).action_id
```
jcoux: Do this only if records

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket_stage.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class HelpdeskTicketStage(models.Model):
+    _inherit = "helpdesk.ticket.stage"
+
+    action_id = fields.Many2one(
+        "ir.actions.server",
+        string="Server Action",
+        domain="[('model_id', '=', 'helpdesk.ticket')]",
+        help="The assigned action will be executed when a ticket is assigned to the stage.",
```
jcoux: Please add which values in passed in context of the action (on model helpdesk.ticket and with ticket ids).

helpdesk_mgmt_stage_server_action/tests/test_helpdesk_ticket_stage.py
```diff
@@ -0,0 +1,64 @@
...
+                "action_id": cls.server_action_helpdesk_ticket.id,
+                "sequence": 2,
+            }
+        )
+
+    def test_helpdesk_ticket_create(self):
+        self.helpdesk_ticket_1 = self.HelpdeskTicket.create(
+            {
+                "name": "Helpdesk Ticket 1",
+                "stage_id": self.helpdesk_ticket_stage_2.id,
```
jcoux, line 48: Please also add a test with creating ticket without stage_id: the user must not be changed.

Do the same with a stage, but no action on it.

helpdesk_mgmt_stage_server_action/tests/test_helpdesk_ticket_stage.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_helpdesk_ticket_write(self):
+        self.helpdesk_ticket_2 = self.HelpdeskTicket.create(
+            {
+                "name": "Helpdesk Ticket 2",
+                "stage_id": self.helpdesk_ticket_stage_1.id,
+                "description": "Helpdesk Ticket Description",
+            }
+        )
+        self.assertNotEqual(self.helpdesk_ticket_2.user_id, self.env.user)
+        self.helpdesk_ticket_2.write({"stage_id": self.helpdesk_ticket_stage_2.id})
```
jcoux, line 71: Please also add a test with writing a ticket without stage_id: the user must not changed.

Do the same with a stage, but no action on it.

Do the same by giving the same stage (with the action) we already have on the ticket.

Please also add a test with several tickets, but a part of them already have the new stage: only tickets on which the stage really changed must have a new user.

jcoux (changes requested):

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket.py
```diff
@@ -0,0 +1,41 @@
...
+                }
+                action.with_context(**context).run()
+        return records
+
+    def write(self, vals):
+        records = "stage_id" in vals and self.filtered(
+            lambda l: l.stage_id.id != vals.get("stage_id")
+        )
+        if records:
+            records = self.filtered(lambda l: l.stage_id.id != vals.get("stage_id"))
```
jcoux: You already defined records previously.
```suggestion
```

jcoux (commented):
A little last change

jcoux (approved):

PR #405, opened 2023-01-02
[14.0][IMP]helpdesk_mgmt_fieldservice: improve all_orders_closed

HekkiMelody:
Improved the compute method for all_orders_closed, and reused the field in other places where the same check was done again manually instead of reusing the field.


HekkiMelody:
#454

PR #409, opened 2023-01-04
[14.0][FIX] helpdesk_mgmt: assigned date empty and creation method to multi

DavidJForgeFlow:
When creating a new ticket with an assigned user, the assigned date field was being left empty. Now when the user is being assigned in the creation, the assigned date is settled with creation_date.
Also changed the create function from @api.model to @api.model_create_multi.

@ForgeFlow

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -134,6 +134,8 @@ class HelpdeskTicket(models.Model):
             vals["number"] = self._prepare_ticket_number(vals)
         if not vals.get("team_id") and vals.get("category_id"):
             vals["team_id"] = self._prepare_team_id(vals)
+        if vals.get("user_id") and not vals.get("assigned_date"):
+            vals["assigned_date"] = fields.Datetime.now()
         return super().create(vals)
 
     def copy(self, default=None):
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -128,15 +128,16 @@ class HelpdeskTicket(models.Model):
     def _creation_subtype(self):
         return self.env.ref("helpdesk_mgmt.hlp_tck_created")
 
-    @api.model
-    def create(self, vals):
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
Please don't do a write after the super, but include the vals on the creation dictionary with `fields.Datetime.now()`. It would be also advisable to convert the create method to batch one.

DavidJForgeFlow:
> Please don't do a write after the super, but include the vals on the creation dictionary with `fields.Datetime.now()`. It would be also advisable to convert the create method to batch one.

Hi Pedro, I've moved the assign to the vals. But what do you mean with batch method? Do you have any example? Thanks BTW

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -134,6 +134,8 @@ def create(self, vals):
             vals["number"] = self._prepare_ticket_number(vals)
         if not vals.get("team_id") and vals.get("category_id"):
             vals["team_id"] = self._prepare_team_id(vals)
+        if vals.get("user_id"):
```
LoisRForgeFlow: and not vals.get("assigned_date")

DavidJForgeFlow: Done!

LoisRForgeFlow (approved):

pedrobaeza:
I mean to use `@api.model_create_multi`. Example: https://github.com/odoo/odoo/blob/5b3e9a657167b0dd28cca6db5841ed97a9ede44f/addons/account/models/account_move.py#L2028

DavidJForgeFlow:
> I mean to use `@api.model_create_multi`. Example: https://github.com/odoo/odoo/blob/5b3e9a657167b0dd28cca6db5841ed97a9ede44f/addons/account/models/account_move.py#L2028

Done!

MiquelRForgeFlow (approved):

pedrobaeza (approved):
/ocabot merge patch

LoisRForgeFlow:
@DavidJForgeFlow can you forward port this?

PR #416, opened 2023-01-11
[13.0][IMP] helpdesk_mgmt: Only show tickets of the user's helpdesk team with "User: Team tickets" permission.

victoralmau:
Only show tickets of the user's helpdesk team with "`User: Team tickets`" permission (related to https://github.com/OCA/helpdesk/pull/395#pullrequestreview-1229370921)

@Tecnativa TT41172

### DIFF
--- a/helpdesk_mgmt/security/helpdesk_security.xml
+++ b/helpdesk_mgmt/security/helpdesk_security.xml
@@ -32,7 +32,7 @@
             <field ref="model_helpdesk_ticket" name="model_id" />
             <field
                 name="domain_force"
-            >['|',('user_id','=',user.id),('user_id','=',False)]</field>
+            >["|", ('user_id', '=', user.id), '&amp;', ('user_id','=',False), ('team_id', 'in', user.helpdesk_team_ids.ids)]</field>
             <field name="groups" eval="[(4, ref('group_helpdesk_user_own'))]" />
         </record>
         <record id="helpdesk_ticket_team_rule" model="ir.rule">
--- a/helpdesk_mgmt/__manifest__.py
+++ b/helpdesk_mgmt/__manifest__.py
@@ -4,7 +4,7 @@
     "name": "Helpdesk Management",
     "summary": """
         Helpdesk""",
-    "version": "13.0.2.1.0",
+    "version": "13.0.2.2.0",
     "license": "AGPL-3",
     "category": "After-Sales",
     "author": "AdaptiveCity, "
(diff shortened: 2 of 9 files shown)


pedrobaeza:
Why not doing this on the record rules, simply expanding them?

victoralmau:
> Why not doing this on the record rules, simply expanding them?

What rule do you suggest?
Changing https://github.com/OCA/helpdesk/blob/13.0/helpdesk_mgmt/security/helpdesk_security.xml#L35 to limit according to helpdesk teams from user would be "incorrect" in some use cases.

pedrobaeza:
Yes

victoralmau:
Changes done, is this approach what you had in mind?

helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -32,15 +32,13 @@
             <field ref="model_helpdesk_ticket" name="model_id" />
             <field
                 name="domain_force"
-            >['|',('user_id','=',user.id),('user_id','=',False)]</field>
+            >[('team_id','in',user.helpdesk_team_ids.ids), '|',('user_id','=',user.id),('user_id','=',False)]</field>
```
pedrobaeza: This should cover also the case for team_id = False, but user_id = uid.

helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -32,15 +32,13 @@
...
             <field name="groups" eval="[(4, ref('group_helpdesk_user_own'))]" />
         </record>
         <record id="helpdesk_ticket_team_rule" model="ir.rule">
             <field name="name">Team Tickets</field>
             <field ref="model_helpdesk_ticket" name="model_id" />
-            <field
-                name="domain_force"
-            >['|',('team_id','in',user.helpdesk_team_ids.ids),('team_id','=',False)]</field>
+            <field name="domain_force">[('team_id','=',False)]</field>
```
pedrobaeza: This should match team_id = False and team_id = your team

pedrobaeza (changes requested):
Include both changes in the migration script

victoralmau:
> Include both changes in the migration script

Changes done.

helpdesk_mgmt/security/helpdesk_security.xml
```diff
@@ -32,7 +32,7 @@
             <field ref="model_helpdesk_ticket" name="model_id" />
             <field
                 name="domain_force"
-            >['|',('user_id','=',user.id),('user_id','=',False)]</field>
+            >["&amp;","&amp;",('team_id','in',user.helpdesk_team_ids.ids),"|",('user_id','=',user.id),('user_id','=',False),"|",('user_id','=',user.id),('team_id','=',False)]</field>
```
pedrobaeza: Don't get complicated. The correct one should be:

```suggestion
            >["|", ('user_id', '=', user.id), '&amp;', ('user_id','=',False), ('team_id', 'in', user.helpdesk_team_ids.ids)]</field>
```

pedrobaeza (approved):
Let's fix the issue:

/ocabot merge nobump

PR #417, opened 2023-01-12
[14.0][FIX] helpdesk_motive: onchange removed in base module.

LoisRForgeFlow:
After https://github.com/OCA/helpdesk/commit/d053593da61541a6c8fb9a33782656d50c43e3e4 `_onchange_dominion_user_id` has been removed in helpdesk_mgmt, so it cannot be extended anymore in helpdesk_motive.

cc @victoralmau @pedrobaeza

### DIFF
--- a/helpdesk_motive/models/helpdesk_ticket.py
+++ b/helpdesk_motive/models/helpdesk_ticket.py
@@ -12,7 +12,6 @@ class HelpdeskTicket(models.Model):
     )
 
     @api.onchange("team_id", "user_id")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think you should preserve the onchange. Just rename the method to be unique (`_onchange_team_user_helpdesk_motive`), and don't call super.

LoisRForgeFlow:
@pedrobaeza yep, that is better. Done!

pedrobaeza (approved):

victoralmau (approved):

PR #275, opened 2021-10-14
[14.0][IMP] helpdesk_mgmt: allow to define default team for a category

PierrickBrun:
On creation, if category if defined but not the team, the team will be deducted from the category's `default_team_id`

PR #420, opened 2023-01-17
[15.0][IMP] helpdesk_mgmt: allow to define default team for a category

victoralmau:
FWP from 14.0: https://github.com/OCA/helpdesk/pull/275

On creation, if category if defined but not the team, the team will be deducted from the category's default_team_id

@Tecnativa


pedrobaeza:
Why do we need this?

victoralmau:
> Why do we need this?

I discovered that this was added in v14 and wasn't in v15, so I've added it.

pedrobaeza:
Yes, but it's a feature we don't need and it seems to be conflicting according comments after the merge, so I would let it non added if not needed.

victoralmau:
Not yet added in v15 due to conflicts explained in v14.

PR #425, opened 2023-01-20
[15.0][MIG] helpdesk_mgmt_timesheet : Migration to 15.0

vincent-hatakeyama:
Rebase of #338

PR #442, opened 2023-03-02
[15.0][ADD] helpdesk_mgmt_timesheet_time_type: Add Time Type to Tickets timesheet

vincent-hatakeyama:
Goes after #425


helpdesk_mgmt_timesheet/__manifest__.py
```diff
@@ -20,6 +20,7 @@
         "helpdesk_mgmt_project",
         "hr_timesheet",
         "project_timesheet_time_control",
+        "hr_timesheet_time_type",
```
pedrobaeza: This must be a separate module adding such feature, not adding dependencies to the base module.

PR #448, opened 2023-03-29
[14.0] helpdesk_mgmt: Forward port multiple 12.0 commits

ypapouin:
See commits for details.
This is a pack multiple commits authored by @eLBati, fixed and rewritten to be compatible with Odoo 14.0.

- Set "e-mail" channel when ticket generated from fetchmail cron task.
- Add a list of attachments in the ticket portal view.
- When a ticket is created by a customer with attachments, an access token is automatically generated.
- Allow a portal user to close its ticket if the final stage is set with `portal_user_can_close`.
- Sequence field added to allow kanban reordering.
- E-mail templates are simplified, they now all use the `mail.mail_notification_light` layout to have a consistent design.
- Send a notification to the partner each time the ticket stage changes.

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -19,14 +19,22 @@ class HelpdeskTicketController(http.Controller):
                 values[field_name] = int(field_value)
             else:
                 values[field_name] = field_value
-        ticket = (
-            http.request.env["helpdesk.ticket"]
-            .sudo()
-            .search([("id", "=", values["ticket_id"])])
-        )
-        ticket.stage_id = values.get("stage_id")
-
-        return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
+        if "ticket_id" in values:
+            ticket = (
+                http.request.env["helpdesk.ticket"]
+                .sudo()
+                .search([("id", "=", values.get("ticket_id"))])
+            )
+            if ticket and "stage_id" in values:
+                stage = (
+                    http.request.env["helpdesk.ticket.stage"]
+                    .sudo()
+                    .search([("id", "=", values.get("stage_id"))])
+                )
+                if stage and stage.portal_user_can_close:
+                    ticket.stage_id = stage
+                return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
+        return werkzeug.utils.redirect("/my/tickets")
 
     def _get_teams(self):
         return (
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -99,10 +99,12 @@ class HelpdeskTicketController(http.Controller):
         new_ticket = request.env["helpdesk.ticket"].sudo().create(vals)
         new_ticket.message_subscribe(partner_ids=request.env.user.partner_id.ids)
         if kw.get("attachment"):
+            IrAttachment = request.env["ir.attachment"]
+            attachment_ids = IrAttachment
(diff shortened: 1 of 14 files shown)


lk-eska:
when do we expect to merge these changes?

helpdesk_mgmt/controllers/main.py
```diff
@@ -90,15 +98,18 @@ def submit_ticket(self, **kw):
         new_ticket = request.env["helpdesk.ticket"].sudo().create(vals)
         new_ticket.message_subscribe(partner_ids=request.env.user.partner_id.ids)
         if kw.get("attachment"):
+            IrAttachment = request.env["ir.attachment"]
+            attachment_ids = IrAttachment
```
HekkiMelody, line 105: what's the point of this?
```suggestion
            attachment_ids = request.env["ir.attachment"]
```

HekkiMelody (approved):
Code review

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -103,8 +105,23 @@ def _read_group_stage_ids(self, stages, domain, order):
...
+        string="Sequence",
+        index=True,
+        default=10,
+        help="Gives the sequence order when displaying a list of tickets.",
+    )
     active = fields.Boolean(default=True)
 
+    def send_partner_mail(self):
+        created_ticket_template = self.env.ref(
+            "helpdesk_mgmt.created_ticket_template", raise_if_not_found=False
```
HekkiMelody, line 122: This template is missing

HekkiMelody (changes requested):
Really want these changes merged.

There are some issues. For example the customer can add some attachments and see them, but there doesn't seem to be a way for the internal operator to upload attachments from the user view and show them for the customer in the portal view. 

Also, I think the "created" email template should not be forcefully sent in the create method, but should be attached to the "new" state and work like every other email template.

That said, I'd rather this be merged and these issues fixes in future PRs, personally. The only thing that's blocking IMO is the missing email template for new tickets.

helpdesk_mgmt/controllers/main.py
ypapouin: `IrAttachment` is used in the create statement below, that's why the model is set once outside the loop

ypapouin:
@aleuffre, the missing template is re-added.

About this, I tried to use the built-in way to send email notification on ticket creation by assigning `helpdesk_ticket_stage_new.mail_template_id` but it's not working when the ticket is created from the portal or duplicated from the backend, I've added a comment in the `send_partner_mail` function (It could be related to the odoo framework).

About attachments, the chatter allows the internal users to add public attachments from the backend (using `Send message`) and since Odoo 14.0, any users (portal/internal) can add public attachments from the portal.
The attachment list is still there as a feedback.

Also, please give me some time, I will try to increase code coverage before merging.

pedrobaeza:
Too much things in one PR. This should be split by each function to be discussed and reviewed isolatedly, or there are few chances of this being merged as is.

ypapouin:
@pedrobaeza, I understand your position but most of these commits are already merged in the 12.0 branch, its like a migration update, others one are tests.

pedrobaeza:
I use the module since 13.0, so I didn't take part on the decisions made on v12, so I'm still reluctant to approve this on batch.

HekkiMelody:
I also need these commits on v14. Will start opening some PRs separately if the whole thing won't be accepted.

ypapouin:
@pedrobaeza as your request, this PR has been split into multiple PRs.
So I'm closing this one.

PR #450, opened 2023-04-18
[16.0][MIG] helpdesk_mgmt_sla

AntoniRomera:


helpdesk_mgmt_sla/views/helpdesk_ticket_team_views.xml
```diff
@@ -7,7 +7,7 @@
         <field name="arch" type="xml">
             <xpath expr="//group[@name='right']" position="inside">
                 <field name="use_sla" />
-                <field name="resource_calendar_id" />
+                <field name="resource_calendar_id" groups="base.group_multi_company" />
```
pedrobaeza, line 10: This is not correct. Add instead the field `company_id` without group as invisible.

pedrobaeza (changes requested):
Please also squash translation commits from the same translator and add the pre-commit stuff in a separate commit.

andreampiovesana:
@hgarridoco can you fix?

CarFla:
Hi.
this mean that this module will not be ported to odoo 16.0??

PR #452, opened 2023-04-18
[16.0][IMP] helpdesk_mgmt

AntoniRomera:
- Visual improvement in the ticket list of the portal
- Quitar el boton de "Crear ticket" de "Portal My  Home"

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
@@ -36,19 +36,6 @@
                 <t t-set="url" t-value="'/my/tickets'" />
                 <t t-set="placeholder_count" t-value="'ticket_count'" />
             </t>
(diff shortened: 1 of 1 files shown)


moitabenfdz:
## In /my/home
### Before:
imagen
### After:
imagen

## In /my/tickets
### Before:
imagen
### After:
imagen

Although it is more elegant, I think it will be more difficult for a non-experienced user to find the button. In my opinion, it could worsen the user experience. I don't think it should be in the middle of the search and filter options.

My proposal would be something like this, but I'm not really convinced.
imagen

AntoniRomera:
> ore difficult for a non-experienced user to find the button. In my opinion, it could worsen the user experience. I don't think it should be in the middle of the search and filter options.

Can we put the button in green? Since in an inexperienced user it is true that it would be difficult to find it, but if we put it with a green color it will be easier and then for experienced users it gives them a cleaner view.

PR #456, opened 2023-04-26
[13.0][FIX] helpdesk_mgmt: portal view

chienandalu:
Portal users can't access to partner's records data so if a users partner is subscribed to another partner ticket we'll get an access error whenever the user tries to get into the portal. We can use in this case the t-field directive to show the info.

cc @Tecnativa TT42618

please review @victoralmau @pedrobaeza

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
@@ -91,7 +91,7 @@
                         <t t-foreach="tickets" t-as="ticket">
                             <tr>
                                 <td>
-                                    <t t-esc="ticket.partner_id.name" />
+                                    <span t-field="ticket.partner_id" />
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think this is a glitch in `name_get`, but this shouldn't work in upper versions. What about putting `sudo` instead in the t directive?

chienandalu:
> I think this is a glitch in `name_get`, but this shouldn't work in upper versions. What about putting `sudo` instead in the t directive?

That's the explicit behavior for this directive when there's no field value: https://github.com/OCA/OCB/blob/1785047d58f28ba11e11e40ba134e89bb235fa8b/odoo/addons/base/models/ir_qweb_fields.py#L316

pedrobaeza (approved):
OK, I see... It's kept across versions, so let's merge it:

/ocabot merge minor

PR #462, opened 2023-04-28
[15.0][FIX] helpdesk_mgmt: portal view

chienandalu:
fw of #461 
Portal users can't access to partner's records data so if a users partner is subscribed to another partner ticket we'll get an access error whenever the user tries to get into the portal. We can use in this case the t-field directive to show the info.

cc @Tecnativa TT42618

ping @victoralmau @pedrobaeza


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #465, opened 2023-05-10
[14.0][FW-PORT] helpdesk_mgmt: Email channel

ypapouin:

- Set "e-mail" channel when ticket generated from fetchmail cron task.

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -121,6 +121,12 @@ class HelpdeskTicket(models.Model):
             self.partner_email = self.partner_id.email
             self.partner_lang = self.partner_id.lang
 
+    def _get_default_email_channel(self):
+        return self.env.ref(
+            "helpdesk_mgmt.helpdesk_ticket_channel_email",
+            raise_if_not_found=False,
+        )
+
     # ---------------------------------------------------
     # CRUD
     # ---------------------------------------------------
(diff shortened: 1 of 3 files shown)


helpdesk_mgmt/tests/common.py
```diff
@@ -33,6 +33,7 @@ def setUpClass(cls):
             context=ctx,
         )
         cls.stage_closed = cls.env.ref("helpdesk_mgmt.helpdesk_ticket_stage_done")
+        cls.channel_email = cls.env.ref("helpdesk_mgmt.helpdesk_ticket_channel_email")
```
pedrobaeza: No need to put this in the base. It's not adding too much value for the rest of the tests.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -137,6 +137,14 @@ def create(self, vals_list):
...
             if vals.get("user_id") and not vals.get("assigned_date"):
                 vals["assigned_date"] = fields.Datetime.now()
+            # Automatically set e-mail channel when created from the
+            # fetchmail cron task
+            if self.env.context.get("fetchmail_cron_running") and not vals.get(
+                "channel_id"
+            ):
+                vals["channel_id"] = self.env.ref(
+                    "helpdesk_mgmt.helpdesk_ticket_channel_email"
+                ).id
```
pedrobaeza: Protect this call with second argument `False` for avoiding an error if the channel is removed.

pedrobaeza (changes requested):

PicchiSeba (commented):
Code review: LGTM

PicchiSeba:
I can't approve this PR, maybe because it's closed

PR #383, opened 2022-11-17
[15.0][IMP] helpdesk_mgmt: Add "close from portal" option & test

houzefa-abba:
By default portal users can close tickets they have submitted; close buttons available to them are all ticket stages marked as "closed".
This PR gives finer control on which stages portal users may switch to when closing a ticket.
Typical use case: a "declined by customer support" stage that marks the ticket as closed but should not be available to portal users.

* Add tests around close buttons available in portal forms + a guard against invalid stage request in the controller
* Add a setting in ticket stage configuration to control which close buttons are available in portal mode:
    * Ticket stage: Add a "close from portal" setting.
    * Portal form: Display ticket close buttons according to that setting.
    * Default stages: Add a "rejected" closed stage unavailable to portal.

Stage config with new setting:
Screenshot at 2022-11-17 11-52-16

Close buttons ("terminé" & "annulé" in French) are now displayed based on that setting:
Screenshot at 2022-11-17 11-52-33

PR #466, opened 2023-05-10
[14.0][FW-PORT] helpdesk_mgmt: Allow portal user to close its tickets

ypapouin:

- Allow a portal user to close its ticket if the final stage is set with `portal_user_can_close`.

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -20,14 +20,22 @@ class HelpdeskTicketController(http.Controller):
                 values[field_name] = int(field_value)
             else:
                 values[field_name] = field_value
-        ticket = (
-            http.request.env["helpdesk.ticket"]
-            .sudo()
-            .search([("id", "=", values["ticket_id"])])
-        )
-        ticket.stage_id = values.get("stage_id")
-
-        return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
+        if "ticket_id" in values:
+            ticket = (
+                http.request.env["helpdesk.ticket"]
+                .sudo()
+                .search([("id", "=", values.get("ticket_id"))])
+            )
+            if ticket and "stage_id" in values:
+                stage = (
+                    http.request.env["helpdesk.ticket.stage"]
+                    .sudo()
+                    .search([("id", "=", values.get("stage_id"))])
+                )
+                if stage and stage.portal_user_can_close:
+                    ticket.stage_id = stage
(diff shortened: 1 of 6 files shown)


helpdesk_mgmt/controllers/main.py
```diff
@@ -11,7 +11,9 @@
 
 
 class HelpdeskTicketController(http.Controller):
-    @http.route("/ticket/close", type="http", auth="user")
+    """ """
```
pedrobaeza, line 14: Unneeded change

helpdesk_mgmt/controllers/main.py
```diff
@@ -20,14 +22,22 @@ def support_ticket_close(self, **kw):
...
                 values[field_name] = field_value
-        ticket = (
-            http.request.env["helpdesk.ticket"]
-            .sudo()
-            .search([("id", "=", values["ticket_id"])])
-        )
-        ticket.stage_id = values.get("stage_id")
-
-        return werkzeug.utils.redirect("/my/ticket/" + str(ticket.id))
+        if "ticket_id" in values:
```
pedrobaeza, line 25: Why is this needed and is it related to the change of this PR?

helpdesk_mgmt/views/helpdesk_ticket_templates.xml
```diff
@@ -136,32 +136,34 @@
...
                             </div>
+                            <div class="col-md-6">
                             <t t-foreach="closed_stages" t-as="stage">
                                 <form
-                                    method="GET"
-                                    t-if="not ticket.closed_date"
-                                    t-attf-action="/ticket/close"
-                                    style="display:inline;"
-                                >
+                                        method="GET"
```
pedrobaeza, line 142: Incorrect extra indentation (for all the rest of the diff)

pedrobaeza (changes requested):
Something similar has been put on 15.0 in #383, so I think it's better to backport such feature to equal modern branches.

helpdesk_mgmt/controllers/main.py
ypapouin: What do you mean, adding the check for `ticket_id` ?

pedrobaeza: Yes, why it was not needed and now it is? And if it's an unrelated patch, it shouldn't go here.

pedrobaeza: Anyway, I think the 15.0 change should be backported instead.

PR #467, opened 2023-05-10
[14.0][FW-PORT] helpdesk_mgmt: Simplify e-mail templates

ypapouin:

- E-mail templates are simplified, they now all use the `mail.mail_notification_light` layout to have a consistent design.
- Send a notification to the partner each time the ticket stage changes.


helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -75,8 +75,8 @@
                 name="body_html"
             ><![CDATA[<html>
                 <div>
-                    <p>Hello ${object.user_id.name},</p>
-                    <p>The ticket "${object.display_name}" stage has changed to ${object.stage_id.name}.</p>
+                    <p>Hello ${object.partner_id.name or ''},</p>
+                    <p>The ticket ${object.number or 'n/a'} stage has changed to: ${object.stage_id.name}.</p>
```
pedrobaeza: I prefer the previous one.

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -102,13 +102,15 @@
             <field name="name">In Progress</field>
             <field name="unattended">False</field>
             <field name="closed">False</field>
+            <field name="mail_template_id" ref="helpdesk_mgmt.changed_stage_template" />
```
pedrobaeza: Don't add this by default. Customize it in your end.

pedrobaeza (changes requested):
This is also mixing several things. All related to mail templates, but too much extreme changes IMO.

helpdesk_mgmt/data/helpdesk_data.xml
ypapouin: The `changed_stage_template` exists but is not used anywhere. For me it is more logical to add it by default and so easier customize it by simply removing if from stages.

pedrobaeza: No, it's not a good default. People not aware of this will have mails sent on any move. It's there, but not filled, because it's optional.

ypapouin:
The `display_name` is now used for all templates body.
I also edited the doc to explain the stage's email template.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -111,6 +113,19 @@ def _read_group_stage_ids(self, stages, domain, order):
     )
     active = fields.Boolean(default=True)
 
+    def send_partner_mail(self):
+        """Send a feedback e-mail to the partner that created the ticket.
+        Note: Assigning `mail_template_id` to `helpdesk_ticket_stage_new` doesn't work
+        when the ticket is created from the portal, that's the purpose of this
+        function."""
+        created_ticket_template = self.env.ref(
```
pedrobaeza: This has to be optional through any configuration. Maybe you can put a configuration setting for selecting the template, with this one as default.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -143,7 +158,13 @@ def create(self, vals_list):
                 vals["team_id"] = self._prepare_team_id(vals)
             if vals.get("user_id") and not vals.get("assigned_date"):
                 vals["assigned_date"] = fields.Datetime.now()
-        return super().create(vals_list)
+        res = super().create(vals_list)
+        for rec in res:
+            if rec.partner_id or rec.partner_email:
+                rec.send_partner_mail()
+                if rec.partner_id:
+                    rec.message_subscribe(partner_ids=rec.partner_id.ids)
```
pedrobaeza: Please check how standard does this without overwriting the create, but using `thread` specific methods.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -318,3 +338,39 @@ def action_do_send_email(self):
             "target": "new",
             "context": ctx,
         }
+
+    # ---------------------------------------------------
+    # Portal
+    # ---------------------------------------------------
+
+    def _compute_access_url(self):
+        super(HelpdeskTicket, self)._compute_access_url()
```
pedrobaeza, line 362: Why calling super?

ypapouin: Because it inherited from `portal.mixin`

helpdesk_mgmt/models/helpdesk_ticket.py
ypapouin: The partner subscribe is now handled using `_message_auto_subscribe_followers` from `thread`, but the mail is still sent from the create method

helpdesk_mgmt/models/helpdesk_ticket.py
ypapouin: There is now a system variable named `helpdesk_mgmt.created_ticket_template_name`.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -219,6 +251,10 @@ def message_new(self, msg, custom_values=None):
             "partner_email": msg.get("from"),
             "partner_id": msg.get("author_id"),
         }
+        res = getaddresses([msg.get("from", "")])
+        if res:
+            defaults["partner_name"] = res[0][0]
+            defaults["partner_email"] = res[0][1]
```
renda-dev: **nitpick:** Maybe you can first check if `msg["from"]` is empty or not and then use `getaddresses`

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -14,68 +14,54 @@
...
+            <field name="name">Helpdesk Created Ticket Notification Email</field>
+            <field name="model_id" ref="helpdesk_mgmt.model_helpdesk_ticket" />
+            <field name="email_from">${object.company_id.partner_id.email}</field>
+            <field
+                name="email_to"
+            >${('' if object.partner_id.email and object.partner_id.email == object.partner_email else object.email) | safe},</field>
+            <field name="subject">The ticket ${object.number} has been created.</field>
+            <field
+                name="partner_to"
+            >${object.partner_id.id if object.partner_id.email and object.partner_id.email == object.partner_email else False}</field>
```
PicchiSeba, line 51: **question**: So the way this works is `partner_to` gets set (but `email_to` does not) if the partner's email has an appropriate value and vice versa, right?

I'm asking to make sure I understood correctly.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -318,3 +354,39 @@ def action_do_send_email(self):
...
+            return False
+        user = (
+            self.env["res.users"]
+            .sudo()
+            .search([("partner_id", "=", self.partner_id.id)])
+        )
+        if not user:
+            return False
+        if not self.with_user(user).check_access_rights("read", raise_exception=False):
+            return False
```
renda-dev: **nitpick:**
```suggestion
        if not user or not self.with_user(user).check_access_rights("read", raise_exception=False):
            return False
```

renda-dev (approved):
Just a couple of quick and really small changes could be made but the code LGTM overall.

helpdesk_mgmt/data/helpdesk_data.xml
ypapouin: Indeed, it mimic the way it is done in official `calendar` module
(https://github.com/odoo/odoo/blob/71db47cff99ce3b2090c3be5dfa3205654625d51/addons/calendar/data/mail_template_data.xml#L9-L10)

PicchiSeba: Thanks.

Fine by me then :)

helpdesk_mgmt/models/helpdesk_ticket.py
ypapouin: Done

PR #415, opened 2023-01-11
[14.0][IMP] helpdesk_mgmt_project: apply default project to tickets from portal.

LoisRForgeFlow:
Also a small fix done in the team feature in the base module. See commits.

@ForgeFlow

PR #468, opened 2023-05-10
[14.0][FIX] helpdesk_mgmt_timesheet: Ticket submit broken

ypapouin:
If `helpdesk_mgmt_project` is installed without `helpdesk_mgmt_timesheet`, an `AttributeError` is raised when submitting a new ticket.

Introduced in #415 
Ping @LoisRForgeFlow


LoisRForgeFlow (approved):
LGTM, thanks for the test case :+1:

LoisRForgeFlow:
cc @pedrobaeza

helpdesk_mgmt/tests/test_helpdesk_portal.py
```diff
@@ -47,30 +46,32 @@ def setUp(self):
...
+            "csrf_token": http.WebRequest.csrf_token(self),
+            "subject": self.new_ticket_title,
+            "description": "\n".join(self.new_ticket_desc_lines),
+        }
+        data.update(**values)
+        resp = self.url_open("/submitted/ticket", data=data)
         self.assertEqual(resp.status_code, 200)
 
+
+@odoo.tests.tagged("post_install", "-at_install")
```
pedrobaeza, line 73: Why this?

pedrobaeza (changes requested):
Are the helpdesk teams something exclusive of `helpdesk_mgmt_project`?

helpdesk_mgmt/tests/test_helpdesk_portal.py
ypapouin: Because the original TestHelpdeskPortal in `helpdesk_mgmt` was already tagged with this, I assumed it was essential for any HttpCase.

pedrobaeza: OK, I see, but why not keeping it in the base class? Isn't working that way?

LoisRForgeFlow: @ypapouin could you have a look at @pedrobaeza last questions? it would be great to get this PR merged

ypapouin: Sorry, I missed the question.
If I remember, I made a quick code audit (odoo/oca) to see how base/common test classes were handled in such cases.
Each time, the base class don't have any decorator, only child classes have.

pedrobaeza: Well, but if it's working, it's better to not increase the diff. Here there's no guideline about this.

ypapouin: Ok I made a test and the decorator is not taken into account if set on the base class.

pedrobaeza: OK then. Good to know. Weird, but if happening, indeed it should be changed.

pedrobaeza:
Please rebase after doing https://github.com/OCA/helpdesk/pull/472

ypapouin:
I will rebase but your PR targets 15.0

pedrobaeza (approved):
OK, if this also fixes in this branch the other problem, one bug less

/ocabot merge minor

PR #481, opened 2023-07-03
[16.0][MIG] helpdesk_mgm_timesheet

AntoniRomera:
Removed dependency and usability "project_timesheet_time_control" because at the moment in version 16.0 does not exist and can not be migrated due to the dependency with the analytical tags that I have not yet seen if OCA will keep or not, at the time it is available "project_timesheet_time_control" I will do the PR to put back the usability to this module.


pedrobaeza:
`project_timesheet_time_control` doesn't depend in anything about the analytic tags... and it's mostly migrated, but that one also can remove several dependencies. Please check.

AntoniRomera:
> `project_timesheet_time_control` doesn't depend in anything about the analytic tags... and it's mostly migrated, but that one also can remove several dependencies. Please check.

My fault, I didn't pay attention, now in the PR I will add the functionality again and we will wait for the dependency.

Issue #435, opened 2023-02-09
Incorrect mail address creation when ticket change state 


pepinho7:
## Module
helpdesk_mgmt

## Describe the bug
Once the ticket has been created via email from an unregistered user and the stage is changed to 'Done', a contact is created with the wrong email. The **&gt** is added at the end of the email address. For this reason the sending of the mail for the change of stage fails.

## To Reproduce
**Affected versions**:

1.  Go to Settings > Incoming mail servers and set in 'Actions to Perform on Incoming Mails' **Helpdesk ticket** for _Create a New Record_
2. Create a Team with a mail alias that can accept emails from everyone
3. Send from a mail to that alias with an unregistered (external) user
4. Move the automatically created ticket to the 'Done' stage

**Expected behavior**
A mail message should be sent to the mail address that generated the ticket

**Additional context**
Running Odoo 14.0
![resend_mail](https://user-images.githubusercontent.com/54631469/217847819-c1a2be58-f7e9-4efc-bdae-c07e52abaf2e.png)

afbenevides:
Hello, I have same problem.

Before installing module named: Email_Format_Checker (partner_email_check), I had an error message on a new created ticket from outside user by mail: Validation error, Email invalid

After installing Email_Format_Checker, I have exact same error as you,  Validation error, external.user@mail.xx&gt is an invalide email

Did you find the solution to that problem?

mathben:
What is the content of the field partner_email?
When a ticket is create by an external email by email gateway, it's create email like '"Name" <name@company.com>". It supposes to be "name@company.com". When close tickets, I have the same error of you.
Check #491

PR #491, opened 2023-07-23
[14.0][FIX] Email formating when receive it by mail gateway, create new ticket with wrong label

mathben:
[FIX] helpdesk_mgmt: format email when receive it from new message mail gateway

Before, email was like '"Name" <name@company.com>'
Now, email is like 'name@company.com'
To field partner_email into model helpdesk.ticket

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -213,12 +213,17 @@ class HelpdeskTicket(models.Model):
         """
         if custom_values is None:
             custom_values = {}
+        partner_name, partner_email = next(
+            iter(tools.email_split_tuples(msg.get("from"))), ("", "")
+        )
         defaults = {
             "name": msg.get("subject") or _("No Subject"),
             "description": msg.get("body"),
-            "partner_email": msg.get("from"),
+            "partner_email": partner_email,
             "partner_id": msg.get("author_id"),
         }
+        if not msg.get("author_id"):
+            defaults["partner_name"] = partner_name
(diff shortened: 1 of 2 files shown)


mathben:
This fix #435

mathben:
What do you think about the coverage, do I need to do more test? It's the method _extract_email I created.

pedrobaeza:
Please check how it's done in `project` module and mimic it.

mathben:
> Please check how it's done in `project` module and mimic it.

This is the code into module project v14.0
https://github.com/odoo/odoo/blob/14.0/addons/project/models/project.py#L1312
I think it has the same problem, I will test it.
`tools.email_split()` this seems the way instead of using regex.

mathben:
> Please check how it's done in `project` module and mimic it.

It works now, I tested it.
helpdesk_mgmt it's a different case of project. It's support when partner_id not exist with partner_email and partner_name. So we can fill partner_name from name of email. I can use method `tools.email_split_tuples()` instead of `tools.email_split()`

Thanks for your reviews.

mathben:
Codecov not pass, because of 
```Python
        if not msg.get("author_id"):
            defaults["partner_name"] = partner_name
```
Do you want me to add a duplicate test without author?
I think I can duplicate test method `test_helpdesk_ticket_message_new` without `"partner_ids": [newPartner.id],`

Thanks

pedrobaeza (approved):

mathben:
I rebase with 14.0, ready to merge!

pedrobaeza:
/ocabot merge patch


Please fw-port it to upper versions.

PR #492, opened 2023-07-23
[14.0][FIX] helpdesk_mgmt: hide false name when send email when change stage

mathben:
When forgot to put name of user in ticket, it send False to all. Replace it by empty string. It's ok to receive email like "Hello , [...]"


helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -45,7 +45,7 @@
                           <table width="590" border="0" cellpadding="0" bgcolor="#ffffff" style="min-width: 590px; background-color: rgb(255, 255, 255); padding: 20px; border-collapse:separate;">
                             <tbody>
                               <td valign="top" style="font-family:Arial,Helvetica,sans-serif; color: #555; font-size: 14px;">
-                                <p>Hello ${object.user_id.name},</p>
+                                <p>Hello ${object.user_id.name and object.user_id.name or ''|safe},</p>
```
pedrobaeza: `safe` part is for emails.

```suggestion
                                <p>Hello ${object.user_id.name and object.user_id.name or ''},</p>
```

Now that we are here, the user is the assigned responsible, but this mail is sent to the customer (object.partner_id), so it's not coherent.

I think we should put the partner's name.

mathben:
I finished and tested, using `object.partner_name` instead of `object.user_id.name`.
Ready for review. Thanks

mathben:
I update, it chooses `object.partner_name` by default, if empty, `object.partner_id.name`, else nothing.
This is a better case? Like that, the name can be overwrite it manually.

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -45,7 +45,7 @@
                           <table width="590" border="0" cellpadding="0" bgcolor="#ffffff" style="min-width: 590px; background-color: rgb(255, 255, 255); padding: 20px; border-collapse:separate;">
                             <tbody>
                               <td valign="top" style="font-family:Arial,Helvetica,sans-serif; color: #555; font-size: 14px;">
-                                <p>Hello ${object.user_id.name},</p>
+                                <p>Hello ${object.partner_name and object.partner_name or (object.partner_id.name and (object.partner_id and object.partner_id.name) or '')},</p>
```
pedrobaeza: No need to repeat parts of the expression or useless ones, and the contact name should have priority (you may have updated it according your casing preferences for example):

```suggestion
                                <p>Hello ${object.partner_id.name or object.partner_name  or '')},</p>
```

(the same for both)

pedrobaeza (approved):

victoralmau (approved):
Will it be necessary to add a migration script with noupdate_changes to update the template and translations?

pedrobaeza:
Good point, Víctor. Can you point to an example so that Mathieu can mimic it?

mathben:
Thanks Víctor, you did all the job to add a migration script with noupdate_changes ;-)
You update helpdesk_mgmt.pot and es.po file too, is it ok?

pedrobaeza:
Yes, let's merge it for avoiding conflicts:

/ocabot merge nobump

PR #495, opened 2023-07-27
[16.0][IMP]helpdesk_mgmt: Improve UI/UX

moitabenfdz:
Delete "New Ticket" button from /my
Change position of "New Ticket" button in /my/tickets

This is a new approach of #452


flotho (approved):
functionnal test ok, 
code review OK,
LGTM

rvalyi (commented):
much better!

tarteo (approved):
Thanks! Much better like this.
If you can improve the form also that would be really much appreciated.

image

Either way lets merge this. @OCA/helpdesk-maintainers

pedrobaeza (changes requested):
The "New ticket" button in /my should be configurable, but you can't rip-off an existing option that will damage those that use it.

PR #508, opened 2023-09-28
[15.0][FIX] helpdesk_mgmt: Fixes to automatic email sending according to stage

victoralmau:
Fixes to automatic email sending according to stage:

- [x] Set `stage_id` in the ticket creation from portal so that `_track_template()` method has the correct behavior.
- [x] Set composition_mode=mass_mail so that the email is always sent (for example if the ticket is created from portal).

@Tecnativa TT45288

### DIFF
--- a/helpdesk_mgmt/controllers/main.py
+++ b/helpdesk_mgmt/controllers/main.py
@@ -75,6 +75,11 @@ class HelpdeskTicketController(http.Controller):
             "partner_id": request.env.user.partner_id.id,
             "partner_name": request.env.user.partner_id.name,
             "partner_email": request.env.user.partner_id.email,
+            # Need to set stage_id so that the _track_template() method is called
+            # and the mail is sent automatically if applicable
+            "stage_id": request.env["helpdesk.ticket"].default_get(["stage_id"])[
+                "stage_id"
+            ],
         }
         if company.helpdesk_mgmt_portal_select_team:
             team = (
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -180,6 +180,8 @@ class HelpdeskTicket(models.Model):
             res["stage_id"] = (
                 ticket.stage_id.mail_template_id,
                 {
+                    # Need to set mass_mail so that the email will always be sent
+                    "composition_mode": "mass_mail",
                     "auto_delete_message": True,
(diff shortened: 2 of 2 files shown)


chienandalu (approved):

helpdesk_mgmt/controllers/main.py
```diff
@@ -75,6 +75,9 @@ def _prepare_submit_ticket_vals(self, **kw):
             "partner_id": request.env.user.partner_id.id,
             "partner_name": request.env.user.partner_id.name,
             "partner_email": request.env.user.partner_id.email,
+            # Need to set stage_id so that the _track_template() method is called
+            # and the is sent automatically if applicable
```
pedrobaeza:
```suggestion
            # and the mail is sent automatically if applicable
```

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -180,6 +180,8 @@ def _track_template(self, tracking):
             res["stage_id"] = (
                 ticket.stage_id.mail_template_id,
                 {
+                    # Need to set mass_mail so that the email will always be sent
+                    "composition_mode": "mass_mail",
```
pedrobaeza, line 184: I'm not sure about this. If you are doing this to have a `mail.mail` record, that's something not needed. `mail.message` is enough.

victoralmau: Example of use case:
- An email template is set in the New stage.
- A ticket is created (from the portal form for example).

Before this change:
- Although the chatter indicated the sending of the email, it was not really sent (the recipient is the author).

After this change:
- The email is always sent to the `partner_id` (that of the user who created the ticket).

helpdesk_mgmt/controllers/main.py
```diff
@@ -75,6 +75,9 @@ def _prepare_submit_ticket_vals(self, **kw):
             "partner_id": request.env.user.partner_id.id,
             "partner_name": request.env.user.partner_id.name,
             "partner_email": request.env.user.partner_id.email,
+            # Need to set stage_id so that the _track_template() method is called
+            # and the mail is sent automatically if applicable
+            "stage_id": request.env["helpdesk.ticket.stage"].search([], limit=1).id,
```
pedrobaeza:
```suggestion
            "stage_id": request.env["helpdesk.ticket.stage"].default_get(["stage_id"])["stage_id"],
```

pedrobaeza (approved):
/ocabot merge patch

PR #509, opened 2023-09-28
[15.0][IMP] helpdesk_mgmt: Add empty option to Team and Category fields to portal ticket form

victoralmau:
Changes done:

- [x] Add empty option to Team and Category fields to portal ticket form.
- [x] Add settings to define required fields (Category and Team)

**Before**
antes

**After**
despues

@Tecnativa TT45297


pedrobaeza:
This should go along with a configuration to force required or not those data.

victoralmau:
Added changes to set fields as required or not, do you think this is the right approach?

helpdesk_mgmt/controllers/main.py
```diff
@@ -47,13 +47,18 @@ def create_new_ticket(self, **kw):
         )
         email = http.request.env.user.email
         name = http.request.env.user.name
+        company = request.env.company
+        team_id_required_field = "helpdesk_mgmt_portal_team_id_required"
+        category_id_required_field = "helpdesk_mgmt_portal_category_id_required"
```
chienandalu: This implementation is a little bit weird, why not getting the value directly in the dict below?

helpdesk_mgmt/controllers/main.py
```diff
@@ -47,13 +47,20 @@ def create_new_ticket(self, **kw):
...
                 "categories": categories,
                 "teams": self._get_teams(),
                 "email": email,
                 "name": name,
+                "ticket_team_id_required": company[
+                    "helpdesk_mgmt_portal_team_id_required"
+                ],
+                "ticket_category_id_required": company[
+                    "helpdesk_mgmt_portal_category_id_required"
+                ],
```
chienandalu: Why not simply:
```suggestion
                "ticket_team_id_required": (
                    company.helpdesk_mgmt_portal_team_id_required
                ),
                "ticket_category_id_required": (
                    company.helpdesk_mgmt_portal_category_id_required
                ),
```

chienandalu (approved):
:smile: :+1:

victoralmau:
Ping @pedrobaeza

pedrobaeza (commented):
The new required should be True by default, as it's the most similar to the previous one (where you should always select one).

pedrobaeza (approved):
/ocabot merge minor

PR #510, opened 2023-10-03
[15.0][FIX] helpdesk_mgmt: Multi-company compatibility

victoralmau:
Changes done:
- [x] Add `company_id` field to search and tree view.
- [x] Add `with_company()` to get the correct values.
- [x] Add `color_picker` widget to tag view.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT45356

PR #511, opened 2023-10-03
[16.0][FIX] helpdesk_mgmt: Fixes to automatic email sending according to stage + Multi-company compatibility + Add empty option to Team and Category fields to portal ticket form

victoralmau:
FWP from 15.0: https://github.com/OCA/helpdesk/pull/508 + https://github.com/OCA/helpdesk/pull/510 + https://github.com/OCA/helpdesk/pull/509

Changes done:

- [x] Set `stage_id` in the ticket creation from portal so that `_track_template()` method has the correct behavior.
- [x] Set composition_mode=mass_mail so that the email is always sent (for example if the ticket is created from portal).
- [x] Add `company_id` field to search and tree view.
- [x] Add `with_company()` to get the correct values.
- [x] Add `color_picker` widget to tag view.
- [x] Set tree views editable.
- [x] Add empty option to Team and Category fields to portal ticket form.
- [x] Add settings to define required fields (Category and Team)

@Tecnativa TT45288


victoralmau:
I think we could change the tree views to editable and remove the form views of Channels, Categories, Stages and Tags, what do you think?

pedrobaeza:
OK

helpdesk_mgmt/views/helpdesk_ticket_category_views.xml
```diff
@@ -24,40 +24,11 @@
             </search>
         </field>
     </record>
-    <record id="view_helpdesk_category_form" model="ir.ui.view">
```
pedrobaeza: You can still keep the form views. They may be used in some contexts.

helpdesk_mgmt/views/helpdesk_ticket_tag_views.xml
```diff
@@ -49,10 +49,10 @@
         <field name="name">helpdesk.ticket.tag.tree</field>
         <field name="model">helpdesk.ticket.tag</field>
         <field name="arch" type="xml">
-            <tree decoration-muted="not active">
+            <tree decoration-muted="not active" editable="bottom">
                 <field name="name" />
                 <field name="color" widget="color_picker" />
-                <field name="active" widget="boolean_toggle" />
+                <field name="active" invisible="1" />
```
pedrobaeza, line 55: Why changing this?

helpdesk_mgmt/views/helpdesk_ticket_templates.xml
```diff
@@ -454,6 +454,7 @@
                             name="team"
                             required="True"
                         >
+                            <option value="" />
                                                                             <t
```
pedrobaeza: Review general indentation of the file.

helpdesk_mgmt/views/helpdesk_ticket_tag_views.xml
victoralmau: I prefer to leave the existing behavior in other tree views of active by selecting the records and using the corresponding action. 

Would you prefer we keep this with `boolean_toggle`?

pedrobaeza: OK, as the inactive are not shown by default.

pedrobaeza (approved):

chienandalu (approved):

PR #524, opened 2023-11-07
[15.0][FIX] helpdesk_mgmt: Avoid malformed domain error + Change to read_group() method

victoralmau:
Avoid malformed domain error + Change to `read_group()` method

`TypeError: 'NewId' object is not iterable` (since https://github.com/odoo/odoo/commit/7fc3ebf0fabe3d87f98d31fccd1a7eb2618f8f5f)

@Tecnativa TT45844

### DIFF
--- a/helpdesk_mgmt/models/res_partner.py
+++ b/helpdesk_mgmt/models/res_partner.py
@@ -23,14 +23,26 @@ class ResPartner(models.Model):
     )
 
     def _compute_helpdesk_ticket_count(self):
+        ticket_data = self.env["helpdesk.ticket"].read_group(
+            [("partner_id", "child_of", self.ids)], ["partner_id"], ["partner_id"]
+        )
+        active_ticket_data = self.env["helpdesk.ticket"].read_group(
+            [("partner_id", "child_of", self.ids), ("stage_id.closed", "=", False)],
+            ["partner_id"],
+            ["partner_id"],
+        )
+        mapped_data = {x["partner_id"][0]: x["partner_id_count"] for x in ticket_data}
+        active_mapped_data = {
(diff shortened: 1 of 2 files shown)


helpdesk_mgmt/models/res_partner.py
```diff
@@ -25,7 +25,7 @@ class ResPartner(models.Model):
     def _compute_helpdesk_ticket_count(self):
         for record in self:
             ticket_ids = self.env["helpdesk.ticket"].search(
```
pedrobaeza: I think this should be converted to `read_group`.

chienandalu (approved):

helpdesk_mgmt/models/res_partner.py
```diff
@@ -23,14 +23,26 @@ class ResPartner(models.Model):
     )
 
     def _compute_helpdesk_ticket_count(self):
+        ticket_data = self.env["helpdesk.ticket"].read_group(
+            [("partner_id", "child_of", self.ids)], ["partner_id"], ["partner_id"]
+        )
+        active_ticket_data = self.env["helpdesk.ticket"].read_group(
+            [("partner_id", "child_of", self.ids), ("stage_id.closed", "=", False)],
+            ["partner_id"],
```
pedrobaeza, line 31: Group instead by `commercial_partner_id`, and you will have all the results at once.

victoralmau: In `helpdesk.ticket` there is no `commercial_partner_id` and you cannot group by `partner_id.commercial_partner_id` 
Do you mean something else?

pedrobaeza: OK, my fault.

pedrobaeza (approved):
/ocabot merge patch

PR #534, opened 2024-01-18
[16.0][IMP] heldespk_mgmt: Helpdesk stages per teams

carolinafernandez-tecnativa:
Include a field in the stages to define the team. If it is empty, the stage will be available to all teams. If you have a team indicated, that stage will only be visible to tickets that belong to that team.

@Tecnativa
TT45710

@pedrobaeza


pedrobaeza (approved):
You can squash the last 2 commits, as the authorship of the simplification has been lost anyway, and in this context of fw-porting, it's not so important to split both.

victoralmau (approved):

PR #547, opened 2024-02-29
[16.0][FIX] helpdesk_mgmt_timesheet: add group on show_time_control

vincent-hatakeyama:
This field is computed by accessing "account.analytic.line", which default helpdesk users have no access to. The field is only used in domains for displayed to group hr_timesheet.group_hr_timesheet_user, so limit the field to that group too.
The group hr_timesheet.group_hr_timesheet_user has already access to "account.analytic.line".

### DIFF
--- a/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
+++ b/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
@@ -54,7 +54,11 @@
                 />
             </field>
             <tree position="inside">
-                <field name="show_time_control" invisible="1" />
+                <field
+                    name="show_time_control"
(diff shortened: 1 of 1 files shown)


dalonsod (approved):
Only code review, it has all sense. Migrating from v15 o v16 security access has slightly changed due to `groups_id` removal, so this error appeared in v16, but shouldn't be present in v15, anyway could you confirm if this mustn't be ported to v15?

vincent-hatakeyama:
I confirm that there is no need to port those change to 15.0.

dalonsod:
Thanks!

pedrobaeza (approved):
/ocabot merge minor

PR #548, opened 2024-02-29
[16.0][IMP] helpdesk_mgmt_project: Set by default the company as the project one in tickets

vincent-hatakeyama:


helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -14,6 +15,15 @@ class HelpdeskTicket(models.Model):
         store=True,
     )
 
+    # Override field
+    company_id = fields.Many2one(
+        default=lambda self: self.env["project.project"]
```
flotho: Hi,
Thanks for the work
prefer to define a method before the filed declaration and call the method

flotho (changes requested):

helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -14,6 +16,33 @@ class HelpdeskTicket(models.Model):
...
+        store=True,
+        readonly=False,
+        required=True,
+        copy=True,
+        default=_default_company_id,
+    )
+
+    @api.depends("project_id.company_id")
+    def _compute_company_id(self):
+        for ticket in self.filtered(lambda ticket: ticket.project_id):
```
flotho: AFICS, filtering here can leads you to loop on less records than available in self.
IMHO it doesn't respect https://www.odoo.com/documentation/16.0/developer/reference/backend/orm.html#computed-fields where it is supposed to browse *all* records.
If I'm wrong sorry for buzzing here.
Regards

flotho (changes requested):

helpdesk_mgmt_project/models/helpdesk_ticket.py
vincent-hatakeyama: Fixed.

flotho (approved):
thanks

flotho:
Hi @vincent-hatakeyama , any chance to rebase ?

vincent-hatakeyama:
> Hi @vincent-hatakeyama , any chance to rebase ?

Done.

flotho:
Hi @vincent-hatakeyama , any chance to rebase or push a smal change to reset the CI ?

vincent-hatakeyama:
> Hi @vincent-hatakeyama , any chance to rebase or push a smal change to reset the CI ?

Just rebased on latest 16.0

helpdesk_mgmt_project/views/helpdesk_ticket_view.xml
```diff
@@ -33,6 +33,11 @@
         <field name="inherit_id" ref="helpdesk_mgmt.ticket_view_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='partner_email']" position="after">
+                <field
```
flotho, line 36: Hi thanks for this PR, could you explain the use of this ?
AFAICR, either the field is not pushed in the view if you're not in the group and if you're member of the gourp, it's invisible. 
my conclusion is that you never display nor use the field. 
Tanks for your next precisions

flotho: ping @vincent-hatakeyama , any chance to share explanations ?

helpdesk_mgmt_project/models/__init__.py
```diff
@@ -1,4 +1 @@
-from . import helpdesk_ticket
-from . import helpdesk_ticket_team
-from . import project
-from . import project_task
+from . import helpdesk_ticket, helpdesk_ticket_team, project, project_task
```
flotho, line 1: strange lint here, 
I think that is not the one produced by odoo-lint, could you check ?

helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -5,7 +5,9 @@ class HelpdeskTicket(models.Model):
 
     _inherit = "helpdesk.ticket"
 
-    project_id = fields.Many2one(string="Project", comodel_name="project.project")
+    project_id = fields.Many2one(
+        "project.project", string="Project", tracking=True, check_company=True
```
flotho, line 9: please keep comodel_name, it ease the readability even if technically it's correct

helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -14,6 +16,34 @@ class HelpdeskTicket(models.Model):
         store=True,
     )
 
+    @api.model
+    def _default_company_id(self):
+        if self._context.get("default_project_id"):
```
flotho, line 21: I think that this is not safe enough.
I've already have a case where a variable in context was setup with None value.
So it was present in the context but not usefull for browsing and throw an execpetion.
Please adapt

helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -14,6 +16,34 @@ class HelpdeskTicket(models.Model):
...
+        store=True,
+        readonly=False,
+        required=True,
+        copy=True,
+        default=_default_company_id,
+    )
+
+    @api.depends("project_id.company_id")
+    def _compute_company_id(self):
+        for ticket in self:
```
flotho, line 43: a compute method MUST return a value for each element.
Your method doesn't respect this constraint. Please adapt

helpdesk_mgmt_project/views/helpdesk_ticket_view.xml
flotho: any chance to elaborate an explanation here ?

flotho (changes requested):

helpdesk_mgmt_project/models/helpdesk_ticket.py
NL66278: If the variable has a None value, `self._context.get("default_project_id")` will evaluate to False, so should not pose a problem.

flotho: Hi @NL66278 , 
Thanks, sadly it's not what I'm experiencing  : 
image

NL66278: That is because None is not an instance of bool. But that is not how the if statement evaluates False like values. In `if x:` the condition will not be satisfied if x is None, False, 0, an empty list, an empty dictionary, an empty string....

PR #553, opened 2024-03-25
[15.0] [FIX] helpdesk_mgmt: company_id should come from team

eLBati:
Otherwise, creating tickets from fetchmail would always set odoobot's company in new tickets, potentially in contrast to team's company

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -31,6 +31,16 @@ class HelpdeskTicket(models.Model):
             ] + search_domain
         return stages.search(search_domain, order=order)
 
+    def _default_team_id(self):
+        team_id = (
+            self.env["helpdesk.ticket.team"]
+            .search([("user_ids", "in", self.env.uid)], limit=1)
+            .id
+        )
+        if not team_id:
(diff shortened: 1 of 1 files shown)


ByteMeAsap:
@eLBati , the company would be undefined in case it is created manually without adding any team

eLBati:
@ByteMeAsap why do you think this would be a problem?
Record rule `helpdesk_ticket_comp_rule` expects tickets without company.

Anyway, I also added `_default_team_id` , please review

Thanks

marcelsavegnago (approved):
LGTM

flotho (approved):
code review, LGTM

rvalyi (approved):

pedrobaeza:
Reverted in #619 

See https://github.com/OCA/helpdesk/pull/618#pullrequestreview-2281927887

@rvalyi you merged just after I request the review of my colleague, which has been working on this question a lot. It's true that the PR has been here a lot of time, but we didn't see it, sorry. But this is breaking some flows that were designed to work that way. @victoralmau maybe you can add some tests for that flows.

eLBati:
Replacing with https://github.com/OCA/helpdesk/pull/622

PR #561, opened 2024-04-17
[15.0][IMP] helpdesk_mgmt: Show error if file size exceeds limit

victoralmau:
Show error if file size exceeds limit

Similar to what already exists on web, an error will be displayed if a file exceeding the limit is selected from the /new/ticket portal page.

Example use case:
- Go to the /new/ticket page and select some file larger than 128Mb.
- An error will be displayed and the selected file will be "_removed_".

The limit is the same as in `web`, and can be changed by defining the `web.max_file_upload_size` config parameter.

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa TT48681


helpdesk_mgmt/static/src/js/new_ticket.esm.js
CarlosRoca13, line 1: The `.esm.js` extension is used for es6 files that uses `import` `export`, so it is no needed here :)

CarlosRoca13 (approved):
Non blocking comment

pedrobaeza (approved):
/ocabot merge minor

PR #562, opened 2024-04-18
[15.0][FIX] helpdesk_mgmt: Make frontend js files translatables

CarlosRoca13:
cc @Tecnativa TT48681

The steps to reproduce the problem:
1. Go to /new/ticket
2. Set the web to other lang than English
3. Upload a file with higher size than 128 Mb

The tex will not be translated

With this changes the module is marked as translatable on frontend and it will be considered when searching for translations.

ping @victoralmau @pedrobaeza


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

sergio-teruel (approved):

PR #560, opened 2024-04-11
[15.0][ADD] helpdesk_mgmt_partner_sequence

pleirb:
I have migrated this module with odoo-module-migrator from version 13 to 15 and after some tests it seems to work correctly.
This is my first Pull Request. I hope I'm doing things right.

PR #564, opened 2024-04-22
[IMP] helpdesk_mgmt_partner_sequence: black, isort, prettier

pleirb:
I used odoo-module-migrator to migrate from 13.0 to 15.0 and then followed the steps proposed by OCA. I hope to be successful this time and that the boot at least approves me. If something goes wrong, I appreciate if someone can help me with the contribution


dalonsod:
Hello @pleirb thanks for your contribution.

Importing commit history is still missing. According to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

image

it seems you're not applying right highlighted command, because commit history was not imported. In fact, you should change a bit that command and replace `origin/14.0` by `origin/13.0` , because you are migrating from v13 (see https://github.com/OCA/helpdesk/pull/560#issuecomment-2048930854). Could you try?

PR #566, opened 2024-04-24
[15.0] [MIG] helpdesk_mgmt_partner_sequence

pleirb:
This is my third attempt. I hope the boot lets me pass this time. I spend very little time on this and try to do what I can to contribute. Sorry, if I generate repeated cases, the administrators can eliminate what I have done wrong. Thanks for understanding

PR #569, opened 2024-05-04
[16.0] [MIG] helpdesk_mgmt_partner_sequence

pleirb:
So that there is a continuity of #566 , I am publishing now for version 16.0


pleirb:
It has no effect

PR #577, opened 2024-05-28
[16.0][IMP] helpdesk_mgmt: Update priority of portal_my_home template

marcelsavegnago:
We can consider another strategy in the future, but for now, I believe that changing the priority of the portal's ticket menu to the "last position" will significantly improve its visual arrangement.

Priority 40
image

Priority 99
image

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
@@ -26,7 +26,7 @@
         id="portal_my_home"
         name="Portal My Home : ticket entries"
         inherit_id="portal.portal_my_home"
-        priority="40"
+        priority="99"
         customize_show="True"
     >
         <xpath expr="//div[hasclass('o_portal_docs')]" position="inside">


pedrobaeza:
Not sure it's a good idea to put always "Tickets" as the last entry, and more on existing installations.

marcelsavegnago:
> Not sure it's a good idea to put always "Tickets" as the last entry, and more on existing installations.

The ticket menu being in the middle of the other menus looks a bit odd as well. We could consider moving the "New Ticket" button to the end of the menus perhaps. Do you have any suggestions?

marcelsavegnago:
And indeed, changing the priority to the end does not prevent another module with a higher priority from moving the "New Ticket" button back to the middle of the screen.

pedrobaeza:
Why not integrating the button inside the row, at the right of "Tickets"?

marcelsavegnago:
> Why not integrating the button inside the row, at the right of "Tickets"?

Maybe... I'll try...

flotho:
hi @marcelsavegnago ,
have  look to my proposal here https://github.com/OCA/helpdesk/pull/586

PR #587, opened 2024-06-04
[14.0][FIX] helpdesk_mgmt: fix customer image on portal

marcelsavegnago:
Corrects the if condition that checks if the customer has an image, in this case it is checking ticket.user_id but the correct one would be ticket.partner_id


Matthwhy (approved):

antoniospneto (approved):

kaynnan (approved):
LGTM

houzefa-abba (approved):

pedrobaeza:
Please fw-port it if applicable:

/ocabot merge minor

PR #588, opened 2024-06-05
[16.0][FIX] helpdesk_mgmt: Tag tests as per recommendations for HttpCase type tests.

blaskurain:
The tests of helpdesk_mgmt derived from HttpCase fail when we run them together with some personal addons. Tagging the tests as `@tagged("post_install", "-at_install")` as recommended [1] by Odoo fixes the issue without any downside.

[1] https://www.odoo.com/documentation/16.0/developer/reference/backend/testing.html#special-tags

### DIFF
--- a/helpdesk_mgmt/__manifest__.py
+++ b/helpdesk_mgmt/__manifest__.py
@@ -4,7 +4,7 @@
     "name": "Helpdesk Management",
     "summary": """
         Helpdesk""",
-    "version": "16.0.2.4.0",
+    "version": "16.0.2.4.1",
     "license": "AGPL-3",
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Totally! Please put a proper commit message according guidelines, including the module name and an initial line a bit shorter, that you can expand later in subsequent lines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

pedrobaeza:
Can you please rebase the branch instead of doing a pull and adding an undesired commit?

flotho (approved):
LGTM
Thanks

PR #590, opened 2024-06-11
[FIX][16.0]helpdesk_mgmt_timesheet: add proper import to trigger the proper closest suggestion

flotho:
Don't really know why but wizards weren't imported in the init file.

Additionnaly make the inherits method to have a single return instead of multiple ones


rvalyi (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #596, opened 2024-07-01
[16.0]helpdesk_mgmt_project:  add ticket default company

flotho:
supersedes https://github.com/OCA/helpdesk/pull/548/


flotho:
Hi @vincent-hatakeyama any chance to have a review here ?

flotho:
ping @DorianMag @JulienMartinez @gaelTorrecillas @GuiPrime

helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -14,6 +16,34 @@ class HelpdeskTicket(models.Model):
...
+    company_id = fields.Many2one(
+        "res.company",
+        string="Company",
+        compute="_compute_company_id",
+        store=True,
+        readonly=False,
+        required=True,
+        copy=True,
+        default=_default_company_id,
+    )
```
dalonsod: Attributes that don't change (like `comodel_name`, `string` or `required` shouldn't be overwritten IMO.

helpdesk_mgmt_project/models/helpdesk_ticket.py
```diff
@@ -14,6 +16,34 @@ class HelpdeskTicket(models.Model):
...
+
+    @api.depends("project_id.company_id")
+    def _compute_company_id(self):
+        for ticket in self:
```
dalonsod, line 40: `filtered` could be used here

dalonsod (changes requested):

helpdesk_mgmt_project/models/helpdesk_ticket.py
flotho: not sure because official document says it [must assign the value](https://www.odoo.com/documentation/16.0/developer/reference/backend/orm.html#computed-fields) . 
The syntax wans't more correct IMHO, so I added one assignation.
Regards

dalonsod: OK, for non-stored fields I agree; for stored ones not sure (and specially for non-store and editable ones), anyway you can check adding some extra test that cover all cases.

PR #603, opened 2024-07-05
Fix translation error: corrected "cominicazione" to "comunicazione"

helderme:
The original translation contained a typo: "cominicazione". It has been corrected to "comunicazione".


pedrobaeza:
You have to use our OCA Weblate system for these things:

https://odoo-community.org/page/translate

PR #606, opened 2024-07-17
 add helpdesk_mgmt  and Arabic translations

ageeb1982:
The helpdesk_mgmt module has been added, along with Arabic translations, in version 17.0.


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

Please check also the red CIs.

PR #608, opened 2024-07-22
[16.0] [MIG] helpdesk_mgmt_fieldservice: Migration to 16.0

anusriNPS:
MIgrating helpdesk_mgmt_fieldservice to 16.0 version


helpdesk_mgmt_fieldservice/views/res_partner.xml
```diff
@@ -3,13 +3,14 @@
...
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="arch" type="xml">
+            <field name="service_location_id" invisible="1" />
             <button name="action_view_helpdesk_tickets" position="attributes">
                 <attribute
                     name="context"
-                >{'default_partner_id': id, 'default_fsm_location_id': service_location_id}</attribute>
+                    groups="fieldservice.group_fsm_user_own"
+                >{'default_partner_id': id}</attribute>
             </button>
```
renda-dev: **question:** these changes are not really clear to me, could you give me some context and explain these?

anusriNPS: With migration 16.0 page, this point was mentioned:
I am not able to use groups_id in views instead need to add groups attribute in the field itself. Hence this change was introduced

 Any view with groups_id on it now has to move such groups to the elements of the view. But now you can put groups attribute in a view field instead of isolating it on a view without fear of an access error. Check it in https://github.com/odoo/odoo/pull/98551.

renda-dev: Okay, looks great! But why the default location has been removed?

anusriNPS: Sorry for delayed response. I shouldn't have removed service_location_id while incorporating "groups" for the button. Hope the latest code looks fine

helpdesk_mgmt_fieldservice/views/res_partner.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="arch" type="xml">
+            <field name="service_location_id" invisible="1" />
+            <button name="action_view_helpdesk_tickets" position="attributes">
+                <attribute name="groups">fieldservice.group_fsm_user_own</attribute>
+                <attribute
+                    name="context"
+                >{'default_partner_id': id, 'default_fsm_location_id': service_location_id}</attribute>
+            </button>
+        </field>
+    </record>
```
renda-dev: **issue:** it seems that in 14.0, this customization was made available only for users with the group `fieldservice.group_fsm_user_own`, while now you are making the whole button available only for the users with that group.

Could you please check and fix the issue to make the functionality the same as it was before?

helpdesk_mgmt_fieldservice/views/res_partner.xml
renda-dev: Perfect, thanks!

helpdesk_mgmt_fieldservice/views/res_partner.xml
anusriNPS: Removed groups_id and for accessing  service_location_id field which can be accessed only with groups "fieldservice.group_fsm_user_own" is added in the parent view with invisible attribute so that functionality would remain the same.

renda-dev: Could you please double check? I feel like that to preserve the full functionality you should:
- Use two buttons, the one already existing and create a new one
- The first should be visible if you have the group, the other one not
- The one visible if you have the group should also have the customization made in this inheritance 

Let me know if everything is clear :)

anusriNPS:
As per latest changes, when user is provided with group for helpdesk and not for fieldservice, below view is observed:
Screenshot from 2024-09-10 10-26-50

Provided user permission:

Screenshot from 2024-09-10 10-27-12

helpdesk_mgmt_fieldservice/views/res_partner.xml
```diff
@@ -1,17 +1,34 @@
...
         <field name="name">res.partner.form.ticket.context</field>
         <field name="model">res.partner</field>
-        <field name="groups_id" eval="[(4, ref('fieldservice.group_fsm_user_own'))]" />
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="arch" type="xml">
-            <button name="action_view_helpdesk_tickets" position="attributes">
+              <button name="action_view_helpdesk_tickets" position="before">
+                 <field name="service_location_id" invisible="1" />
+              </button>
+              <button name="action_view_helpdesk_tickets" position="attributes">
```
HekkiMelody: **chore**: an issue that we should fix is that the `<button name="action_view_helpdesk_tickets">` is not present in `base.view_partner_form` but is added by `helpdesk_mgmt.view_partner_form` so we should inherit from that view

helpdesk_mgmt_fieldservice/views/res_partner.xml
HekkiMelody: This seems solved.

HekkiMelody (approved):
Minor comment to fix a bug that was already present in previous versions

But in general, LGTM (Code review)

helpdesk_mgmt_fieldservice/views/res_partner.xml
anusriNPS: Yes, this has to be inherited from helpdesk_mgmt.view_partner_form which is the right way to inherit as mentioned by you. 

However, I see issue still persists with updating button with group attribute to "fieldservice.group_fsm_user_own" and having helpdesk_mgmt and fieldservice modules installed along with helpdesk_mgmt_fieldservice with user provided only with helpdesk_mgmt but not with fieldservice, helpdesk button is not visible in the contacts view as shown below:

image

image

Please let me know your views on it.

anusriNPS:
action_view_helpdesk_ticket function is overridden by replacing context when both groups "helpdesk_mgmt.group_helpdesk_user_own" and "fieldservice.group_fsm_user_own" are available 

Functionality for button in 14 and 16 versions:

Odoo 14:
========
Helpdesk - Manager, Fieldservice - Manager - Has context with default_location_id, default_partner_id
Helpdesk - User, Fieldservice - Disable -  Button is visible but context has search_default_open, default_partner_id
Helpdesk - disable, Fieldservice - User only - Button not visible context could not be checked

Odoo 16:
=======
Helpdesk - Manager, Fieldservice - Manager - context contains default_location_id, default_partner_id
Helpdesk - User, Fieldservice - Disable -  Button is visible but context has only search_default_open, default_partner_id
Helpdesk - disable, Fieldservice - User only - Button not visible context could not be checked

helpdesk_mgmt_fieldservice/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def action_view_helpdesk_tickets(self):
+        if self.env.user.has_group(
+            "fieldservice.group_fsm_user_own"
+        ) and self.env.user.has_group("helpdesk_mgmt.group_helpdesk_user_own"):
+            context = dict(self.env.context)
+            context.pop("search_default_open", None)
```
aleuffre, line 12: **question(non-blocking)**: what is the purpose of this line?

HekkiMelody (approved):
Code review, LGTM

PicchiSeba (approved):
Code review: LGTM

imlopes (approved):

marcelsavegnago (approved):
LGTM

HekkiMelody:
@OCA/helpdesk-maintainers Can this be merged? Thanks

renda-dev (approved):

helpdesk_mgmt_fieldservice/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+            self = self.with_context(
+                context,
+                default_partner_id=self.id,
+                default_fsm_location_id=self.service_location_id.id,
+            )
+            return super(ResPartner, self).action_view_helpdesk_tickets()
+        return super().action_view_helpdesk_tickets
```
ivantodorovich: missing () here, to call the super function

```
return super().action_view_helpdesk_tickets()
```

ivantodorovich (changes requested):

helpdesk_mgmt_fieldservice/models/res_partner.py
HekkiMelody: Nice catch, thanks

ivantodorovich: In fact it was @imlopes who caught it, I got it from his 17.0 migration PR: 
- https://github.com/OCA/helpdesk/pull/652

kaynnan (approved):

helpdesk_mgmt_fieldservice/models/res_partner.py
anusriNPS: Updated as suggested

WesleyOliveira98 (approved):

ivantodorovich (approved):

marcelsavegnago:
@anusriNPS can you rebase and check errors ?

SirPyTech:
> @anusriNPS can you rebase and check errors ?

@marcelsavegnago I have added a commit to fix the tests, I had to change another module's code; let me know if that's fine or if you have any suggestions

CristianoMafraJunior (approved):
LGTM

PR #611, opened 2024-07-23
[16.0][IMP] helpdesk_mgmt_project: Multi-company compatibility

victoralmau:
Multi-company compatibility

Example of use case with `helpdesk_mgmt_timesheet`:

- Create a team with no company defined.
- Create a project A for company A.
- Create a project B for company A.
- [Company A] In the created team we define the project A.
- [Company B] In the created team we define project B
- [Company A] We create a ticket and assign the team, project A will be defined.
- [Company B] We create a ticket and assign the team, project B will be defined.

@Tecnativa TT50251

### DIFF
--- a/helpdesk_mgmt_project/migrations/16.0.2.2.0/post-migration.py
+++ b/helpdesk_mgmt_project/migrations/16.0.2.2.0/post-migration.py
@@ -1,0 +1,20 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Set the default_project_id value in the project company."""
+    env.cr.execute(
+        """
+        SELECT htt.id, htt.old_default_project_id, pp.company_id
+        FROM helpdesk_ticket_team htt
+        JOIN project_project pp ON htt.old_default_project_id = pp.id
+        WHERE htt.old_default_project_id IS NOT NULL
+        """
+    )
+    for team_id, old_default_project_id, company_id in env.cr.fetchall():
+        team = env["helpdesk.ticket.team"].with_company(company_id).browse(team_id)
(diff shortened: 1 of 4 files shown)


helpdesk_mgmt_project/migrations/16.0.2.2.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Create a column with the old data and we fill it in."""
+    openupgrade.logged_query(
```
pedrobaeza: Use `copy_columns` instead.

pedrobaeza: Anyway, I think you don't need to copy it. Just rename it, as now data goes to `ir_property` table.

pedrobaeza: You can use `rename_columns` openupgradelib method.

helpdesk_mgmt_project/migrations/16.0.2.2.0/post-migration.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Set the default_project_id value in the project company."""
+    env.cr.execute(
+        """
+        SELECT team.id, team.old_default_project_id, project.company_id
+        FROM helpdesk_ticket_team AS team
+        JOIN project_project AS project ON team.old_default_project_id = project.id
```
pedrobaeza: Use normalized abbreviations from table names for homogeneity.

```suggestion
        FROM helpdesk_ticket_team htt
        JOIN project_project pp ON htt.old_default_project_id = pp.id
```

pedrobaeza (approved):
/ocabot merge nobump

PR #613, opened 2024-07-29
[16.0][IMP] helpdesk_mgmt_timesheet: Show only users for company selected on ticket

carolinafernandez-tecnativa:
- Before this improvement, when you have multi-company environment and a user wants to add new timesheet, you are able to select all users for all companies that you have access to.
- This PR restrict the users that belongs to company where ticket is created.

@Tecnativa
@pedrobaeza 
TT50251

### DIFF
--- a/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
+++ b/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
@@ -147,6 +147,7 @@
                             <field name="date_time" string="Date" required="1" />
                             <field
                                 name="user_id"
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza (changes requested):
Uhm, wait a moment, what you have to restrict is the employee, not the user.

carolinafernandez-tecnativa:
> Uhm, wait a moment, what you have to restrict is the employee, not the user.

employee is not selectable in timesheet, so i close PR as it cannot be avoided.

PR #604, opened 2024-07-09
[16.0][FIX] helpdesk_mgmt: Do not assign company_id null when creating a ticket

victoralmau:
Do not assign company_id null when creating a ticket

Related to https://github.com/OCA/helpdesk/commit/84353f724df47ff58afc22e64bf71722156f6679

Ping @pedrobaeza can you review it?

@Tecnativa TT50031

PR #557, opened 2024-04-03
[16.0] [FIX] helpdesk_mgmt: use company of team on tickets when team is defined

ByteMeAsap:
Fixes https://github.com/OCA/helpdesk/issues/543

PR #618, opened 2024-09-04
[16.0] [FIX] helpdesk_mgmt: company_id should come from team 

LorenzoC0:
Same fix as #553 , replicated in 16.0

> Otherwise, creating tickets from fetchmail would always set odoobot's company in new tickets, potentially in contrast to team's company

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -38,6 +38,16 @@ class HelpdeskTicket(models.Model):
             ] + search_domain
         return stages.search(search_domain, order=order)
 
+    def _default_team_id(self):
+        team_id = (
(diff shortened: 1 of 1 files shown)


marcelsavegnago (approved):

rvalyi (approved):

victoralmau (changes requested):
I do not agree with this change (also not with https://github.com/OCA/helpdesk/pull/553), perhaps a compute store. Lets NOT forget that teams have a _company_id_ field **not required**.

ByteMeAsap:
@LorenzoC0 , this was already fixed with https://github.com/OCA/helpdesk/pull/557

victoralmau:
> @LorenzoC0 , this was already fixed with #557

Exactly, and fixed after with https://github.com/OCA/helpdesk/pull/604

PR #631, opened 2024-09-27
[16.0][FIX] helpdesk_mgmt_timesheet: Apply the required attrs to the project_id field which has groups

victoralmau:
Apply the required attrs to the `project_id` field which has groups

despues

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa


pedrobaeza:
IMO what you should do is to swap the order on the `helpdesk_mgmt_project` instead, and a guideline must be to put always the visible field in first place.

victoralmau:
> IMO what you should do is to swap the order on the `helpdesk_mgmt_project` instead, and a guideline must be to put always the visible field in first place.

Sorry for not clarifying this earlier. The change you suggest does not work (although it applies the `attrs` to the corresponding field, it does not indicate by interface that it is required), for that reason I have done this change.

pedrobaeza (approved):
OK then

PR #632, opened 2024-09-27
[15.0][IMP] helpdesk_mgmt_sale: Allow to assign multiple tickets to an order

pilarvargas-tecnativa:
cc @Tecnativa TT50683

### DIFF
--- a/helpdesk_mgmt_sale/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt_sale/models/helpdesk_ticket.py
@@ -4,7 +4,7 @@ from odoo import api, fields, models
 class HelpdeskTicket(models.Model):
     _inherit = "helpdesk.ticket"
 
-    sale_order_ids = fields.One2many("sale.order", "ticket_id")
+    sale_order_ids = fields.Many2many("sale.order")
     so_count = fields.Integer(
         string="Sale Order Count", compute="_compute_so_count", store=True
     )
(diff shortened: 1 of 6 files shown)


pilarvargas-tecnativa:
@pedrobaeza Do I include migration script?

pedrobaeza:
Yes, please

helpdesk_mgmt_sale/migrations/15.0.1.0.0/post-migration.py
pedrobaeza, line 1: Incorrect migration folder. You should put a higher one, and bump the version in the manifest.

helpdesk_mgmt_sale/models/helpdesk_ticket.py
```diff
@@ -17,9 +17,9 @@ def _compute_so_count(self):
     def action_view_sale_orders(self):
         """Returns action to view sale orders related to this ticket."""
         action = self.env["ir.actions.actions"]._for_xml_id("sale.action_orders")
-        action["domain"] = [("ticket_id", "=", self.id)]
+        action["domain"] = [("ticket_ids", "in", [self.id])]
         action["context"] = {
-            "default_ticket_id": self.id,
+            "default_ticket_ids": [(6, 0, [self.id])],
```
pedrobaeza: I think this won't work

pilarvargas-tecnativa: why not?
ticket

pedrobaeza: For not losing existing data:

```suggestion
            "default_ticket_ids": [(4, self.id)],
```

pilarvargas-tecnativa: Actually with the tests I did, although the context is transmitted to the view it is only affecting when creating a new order and not the existing ones but you are right, the right way is to make sure that the existing ones are not overwritten. The change is done, thanks

pedrobaeza (approved):

victoralmau (approved):

PR #633, opened 2024-10-01
[15.0][IMP] helpdesk_mgmt_sale: Access tickets from the order

pilarvargas-tecnativa:
With the widget ‘many2many_tags’ it is not possible to click on a ticket in the order and access the form. By adding a smartbutton it is possible to access the list of tickets in the order and also consult the form for each one of them.

cc @Tecnativa TT50683

@pedrobaeza @victoralmau please review

### DIFF
--- a/helpdesk_mgmt_sale/models/sale_order.py
+++ b/helpdesk_mgmt_sale/models/sale_order.py
@@ -1,7 +1,15 @@
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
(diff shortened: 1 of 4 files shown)


pedrobaeza (commented):
I think you should keep both, because if not, you are not able to link an existing one, isn't it?

pilarvargas-tecnativa:
> I think you should keep both, because if not, you are not able to link an existing one, isn't it?

Good, thanks!

pedrobaeza (approved):

victoralmau (approved):

PR #529, opened 2024-01-03
[17.0] helpdesk_mgmt

chandni-serpentcs:

PR #638, opened 2024-10-11
[17.0][MIG] helpdesk_mgmt

giarve:
If original author is not available, superseeding #529

anybody wants to create a new icon for 17.0 style? some svg you could use https://www.svgrepo.com/svg/378347/system-help


imlopes (approved):
Thanks for this work ;)

giarve:
@pedrobaeza should I rebase hard on latest 16.0 branch to remove this pandoc stuff from here? I read somewhere that's what you recommended, but I do not know if it is a must.

pedrobaeza:
Do a interactive rebase over 17.0 (not 16.0), and edit the pre-commit commit, removing the .deb, and the continue the rebasing.

ivantodorovich (approved):

imlopes:
@pedrobaeza it's ok for merge this PR ?
Thanks

pedrobaeza (changes requested):
Why are you changing all the XML-IDs? That is only making difficult the migration of all the rest of the dependent modules, and some of the changes are incorrect.

For example, for views, it's correct to start the XML-ID saying the model (`view`), then an identifier, and finish with the type of view (for example, `_form`). Putting `view` at the end is what is incorrect.

imlopes:
> Why are you changing all the XML-IDs? That is only making difficult the migration of all the rest of the dependent modules, and some of the changes are incorrect.
> 
> For example, for views, it's correct to start the XML-ID saying the model (`view`), then an identifier, and finish with the type of view (for example, `_form`). Putting `view` at the end is what is incorrect.

@pedrobaeza can you give an example ?
I checked and on the previous version with have the `_view` at the end : 
https://github.com/OCA/helpdesk/blob/0f8965853181d223156df236b8aec54d8c6f8efa/helpdesk_mgmt/views/helpdesk_dashboard_views.xml#L3

But on this PR, we have the view type : 
https://github.com/OCA/helpdesk/pull/638/files#diff-346c30f602ec791c5cec98db55dd28a4446400acb877323ee25536e72eaa4924R3

Am I missing something ?

pedrobaeza:
Just look the diff of the migration commit:

imagen

Even if the previous XML-ID was not correct (which is more than the new one IMO), changing them arbitrarily is not good for avoiding errors or more work on dependent modules.

imlopes:
Ahh I got it.
You're right.

giarve:
yep. Demo XML-ids are okay to change I imagine? as they do not affect migrations. Otherwise there are some problems with tests and we can use this PR to suffix them according to the OCA guidelines at this point.

the rest of xml-ids are indeed unnecesary to change. they were in the original PR and now I went one by one reverting the changes.

imlopes:
@pedrobaeza it seems ok to merge now. What you think @giarve ?

helpdesk_mgmt/views/helpdesk_dashboard_views.xml
```diff
@@ -100,39 +113,6 @@
                                         </div>
                                     </div>
                                 </div>
-                                <div
```
pedrobaeza: Why are you removing this?

helpdesk_mgmt/views/helpdesk_dashboard_views.xml
```diff
@@ -25,15 +47,6 @@
                                         </a>
                                     </div>
                                 </div>
-                                <div class="o_kanban_manage_button_section">
```
pedrobaeza: Why removing this?

helpdesk_mgmt/demo/helpdesk_demo.xml
```diff
@@ -5,37 +5,28 @@
         <field name="groups_id" eval="[(4,ref('group_helpdesk_user'))]" />
     </record>
     <!-- Categories -->
-    <record id="helpdesk_category_1" model="helpdesk.ticket.category">
+    <record id="helpdesk_category_1_demo" model="helpdesk.ticket.category">
```
pedrobaeza: Don't do these XML-IDs changes. They only add noise and it's not so relevant. It's the same as views.

helpdesk_mgmt/views/helpdesk_dashboard_views.xml
giarve: image

broken dropdown and at the wrong place otherwise.

helpdesk_mgmt/demo/helpdesk_demo.xml
giarve: Fine, makes sense.

helpdesk_mgmt/views/helpdesk_dashboard_views.xml
giarve: hand in hand with the other comment (also related to the dropdown)

helpdesk_mgmt/views/helpdesk_dashboard_views.xml
pedrobaeza: OK, but is it restored then in the proper place?

giarve: yes, same file: https://github.com/OCA/helpdesk/pull/638/commits/ab81fb549352998317eaf948fdfb89e5027ff71a#diff-346c30f602ec791c5cec98db55dd28a4446400acb877323ee25536e72eaa4924R15

pedrobaeza (approved):
Thanks for the changes

/ocabot merge nobump

PR #639, opened 2024-10-11
[MIG] helpdesk_mgmt_sla: Migration to 16.0

LorenzoC0:
Based off #601 

I've noticed an old PR was closed #450, but I don't understand the change request made by @pedrobaeza. Just for completeness I've already included it in this version.


pedrobaeza:
This claim is legit for not doing incorrect things:

https://github.com/OCA/helpdesk/pull/450#discussion_r1301184170

so if you are repeating the same, it's not correct.

LorenzoC0:
Hi @pedrobaeza ,
maybe I misunderstood your comment, but I removed the domain for the company_id field permanently, and wondered why it was necessary. I noticed that on Odoo core the record rule for multi-company on resource.calendar model is missing. At this point I don't think it's up to this module to add such a feature, rather it would be to create such record rule on core.
What do you think?

pedrobaeza:
What I'm just saying is that when a field protected by a groups is used in a domain, you must add the field 2 times: one with the group, and another invisible without group for not having the error. Simply that.

You don't have to remove any domain.

eLBati:
Moving to https://github.com/OCA/helpdesk/pull/793 addressing the domain problem

PR #643, opened 2024-10-21
[16.0][IMP]helpdesk_mgmt: email channel

PicchiSeba:
Forward-ports https://github.com/OCA/helpdesk/pull/465 which itself forward-ported the commit https://github.com/OCA/helpdesk/commit/fcf0a41bbf4099d7926cf542402d23970a3dcf39 from 12.0 to 14.0

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -165,6 +165,17 @@ class HelpdeskTicket(models.Model):
                 team = self.env["helpdesk.ticket.team"].browse([vals["team_id"]])
                 if team.company_id:
                     vals["company_id"] = team.company_id.id
+            # Automatically set default e-mail channel when created from the
+            # fetchmail cron task
+            if self.env.context.get("fetchmail_cron_running") and not vals.get(
+                "channel_id"
+            ):
+                channel_email_id = self.env.ref(
+                    "helpdesk_mgmt.helpdesk_ticket_channel_email",
+                    raise_if_not_found=False,
+                )
+                if channel_email_id:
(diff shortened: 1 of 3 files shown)


helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -147,6 +147,12 @@ def _onchange_partner_id(self):
             self.partner_name = self.partner_id.name
             self.partner_email = self.partner_id.email
 
+    def _get_default_email_channel(self):
```
pedrobaeza: Why creating this extra method?

PicchiSeba: I can't say for sure as I'm not the original author, but my guess would be that this method was made to be overridable for custom email channels

pedrobaeza: But the way to do that is with default values in the alias, and the other method is already prepared for that, so better to remove it. Let's not bloat the module.

PicchiSeba: If you deem it unnecessary I can remove it

pedrobaeza: Yes

pedrobaeza: It's not removed yet.

PicchiSeba: Yeah I'm sorry, I had just removed its use but not the method itself

flotho (approved):
LGTM, make sense

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please fw-port it to 17

PR #644, opened 2024-10-21
[16.0][IMP]helpdesk_mgmt: simplify email templates

PicchiSeba:
Forward-ports https://github.com/OCA/helpdesk/pull/467 which itself partially forward-ported 98e5ef7f4a2e074283da3674654f662a28f02eed and dfbc3698e76b3a1330f339ca78b49557b78c05ed from 12.0 to 14.0

### DIFF
--- a/helpdesk_mgmt/data/helpdesk_data.xml
+++ b/helpdesk_mgmt/data/helpdesk_data.xml
@@ -32,105 +32,33 @@
             <field name="model_id" ref="helpdesk_mgmt.model_helpdesk_ticket" />
             <field name="email_from">{{object.company_id.partner_id.email}}</field>
             <field
-                name="email_cc"
+                name="email_to"
             >{{not object.partner_id and object.partner_email or ''}},</field>
             <field name="subject">The ticket {{object.number}} has been closed.</field>
             <field name="partner_to">{{object.partner_id.id}}</field>
             <field name="auto_delete" eval="False" />
             <field name="lang">{{object.partner_id.lang}}</field>
             <field name="body_html" type="html">
-                <table
-                    border="0"
-                    width="100%"
-                    cellpadding="0"
-                    bgcolor="#ededed"
-                    style="padding: 20px; background-color: #ededed; border-collapse:separate;"
-                >
-                    <tbody>
-                      <!-- HEADER -->
-                      <tr>
-                        <td align="center" style="min-width: 590px;">
-                          <table
-                                    width="590"
-                                    border="0"
-                                    cellpadding="0"
-                                    bgcolor="#875A7B"
-                                    style="min-width: 590px; background-color: rgb(135,90,123); padding: 20px; border-collapse:separate;"
-                                >
-                            <tr>
-                              <td valign="middle" align="right">
-                                <img
-                                                t-att-src="'/logo.png?company=%s' % object.company_id.id"
-                                                style="padding: 0px; margin: 0px; height: auto; width: 80px;"
-                                                t-att-alt="'%s' % object.company_id.name"
-                                            />
-                              </td>
-                            </tr>
-                          </table>
-                        </td>
-                      </tr>
-                      <!-- CONTENT -->
-                      <tr>
-                        <td align="center" style="min-width: 590px;">
-                          <table
-                                    width="590"
-                                    border="0"
-                                    cellpadding="0"
-                                    bgcolor="#ffffff"
-                                    style="min-width: 590px; background-color: rgb(255, 255, 255); padding: 20px; border-collapse:separate;"
-                                >
-                            <tbody>
-                              <td
-                                            valign="top"
-                                            style="font-family:Arial,Helvetica,sans-serif; color: #555; font-size: 14px;"
-                                        >
-                                <p>Hello <t t-out="object.user_id.name" />,</p>
-                                <p>The ticket "<t
-                                                    t-out="object.display_name"
-                                                />" has been closed.</p>
-                              </td>
-                            </tbody>
-                          </table>
-                        </td>
-                      </tr>
-                      <!-- FOOTER -->
-                      <tr>
-                        <td align="center" style="min-width: 590px;">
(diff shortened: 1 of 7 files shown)


PicchiSeba:
I just rebased because of a conflict arising from the merge of #643

helpdesk_mgmt/controllers/main.py
```diff
@@ -82,6 +82,7 @@ def _prepare_submit_ticket_vals(self, **kw):
             "channel_id": request.env.ref(
                 "helpdesk_mgmt.helpdesk_ticket_channel_web", False
             ).id,
+            "stage_id": request.env.ref("helpdesk_mgmt.helpdesk_ticket_stage_new").id,
```
pedrobaeza: We shouldn't force the initial stage.

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -32,105 +32,53 @@
...
-                                <t t-out="object.company_id.website" />
-                              </td>
-                            </tr>
-                          </table>
-                        </td>
-                      </tr>
-                    </tbody>
-                </table>
+                <div>
+                    <p>Hello ${object.partner_id.name or object.partner_name or ''},</p>
```
pedrobaeza: This is not 16.0 syntax

helpdesk_mgmt/data/helpdesk_data.xml
```diff
@@ -32,105 +32,53 @@
...
+                                View Ticket
+                            </a>
+                        </div>
+                    % endif
+                    Thank you,<br /><br />
+                    ${user.signature | safe}<br />
+                </div>
+            </field>
+        </record>
+        <record id="created_ticket_template" model="mail.template">
```
pedrobaeza: This is a new template.

pedrobaeza (changes requested):
There's an unrelated commit, and the fw-port seems done without checking.

helpdesk_mgmt/data/helpdesk_data.xml
PicchiSeba: Would you prefer its own commit/PR?

pedrobaeza: Or its own PR explaining the reasons, so that not everything must be approved together. Anyway, this is not working, as the syntax is not v16. Did you check it?

PicchiSeba: yes, I'm working on it.

PicchiSeba:
I removed the unrelated commit, the new email template (will do another PR for that) and the `stage_id` forced assignment in the controller. I also converted the email templates to the `16.0` format.

Do you see any other critical issue?

helpdesk_mgmt/tests/test_helpdesk_ticket.py
```diff
@@ -181,3 +183,101 @@ def test_ticket_without_team(self):
...
+                "description": "Ticket test",
+            }
+        )
+        self.assertEqual(t.access_url, "/my/ticket/%s" % t.id)
+        t.partner_id = self.user_portal.partner_id.id
+        self.assertTrue(t.partner_can_access())
+
+    def test_partner_can_access(self):
+        Ticket = self.env["helpdesk.ticket"]
+        self.assertFalse(Ticket.partner_can_access())
```
HekkiMelody: **comment (non-blocking)**: This assertion is a bit redundant IMO

helpdesk_mgmt/tests/test_helpdesk_ticket.py
```diff
@@ -181,3 +183,101 @@ def test_ticket_without_team(self):
             }
         )
         self.assertEqual(self.new_stage, new_ticket.stage_id)
+
+    def test_helpdesk_ticket_multiple(self):
+        Ticket = self.env["helpdesk.ticket"]
+        Partner = self.env["res.partner"]
```
HekkiMelody: **suggestion (non-blocking)**: All these could be put in the setUp once, instead of being repeated in every test.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -176,7 +207,13 @@ def create(self, vals_list):
...
                 if channel_email_id:
                     vals["channel_id"] = channel_email_id.id
-        return super().create(vals_list)
+        res = super().create(vals_list)
+        created_ticket_template_name = self._get_created_ticket_template_name()
+        if created_ticket_template_name:
+            for rec in res:
+                if rec.partner_id or rec.partner_email:
+                    rec._send_mail_to_partner(created_ticket_template_name)
+        return res
```
HekkiMelody: **question**: if you have an email template set on the stage, and this global email template on ticket creation, do both emails get sent at the same time?

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -132,6 +132,28 @@ def _read_group_stage_ids(self, stages, domain, order):
     )
     active = fields.Boolean(default=True)
 
+    def _send_mail_to_partner(self, template_name):
+        """Send a feedback message to the contact linked to this newly created ticket
+        (whatever it is a real partner or just its e-mail).
+        Note: The purpose of this function is to replace the stage's `mail_template_id`
+        role (that just doesn't work when the ticket is created from the portal or when
+        `partner_id` is set)"""
```
HekkiMelody: **issue**: as far as I can tell, so far this doesn't replace but rather adds a new email, no? (see also the other comment below)

HekkiMelody (changes requested):

HekkiMelody:
I see 3 different purposes to this PR
- Refactoring email templates (LGTM on that front)
- Improving how access to tickets is handled for portal users (LGTM on that front too, and it's closely related)
- Hardcoding sending an email on ticket creation no matter what (I have some question, and perhaps this could be split in a different PR?)

PR #650, opened 2024-10-29
[17.0][MIG] helpdesk_mgmt_stage_server_action

yankinmax:


helpdesk_mgmt_stage_server_action/models/helpdesk_ticket.py
```diff
@@ -10,31 +10,22 @@ class HelpdeskTicket(models.Model):
...
-            if action:
-                context = {
-                    "active_model": self._name,
-                    "active_ids": records.ids,
-                }
-                action.with_context(**context).run()
-        else:
-            res = super().write(vals)
+        res = super().write(vals)
+        if "stage_id" in vals:
```
ivantodorovich: The previous method checked if the stage actually changed:

```
records = "stage_id" in vals and self.filtered(
            lambda l: l.stage_id.id != vals.get("stage_id")
```

That seems to have been lost in the refactor.
The unit test checking for this (`test_helpdesk_ticket_write`) was removed, too

This is change in behavior that doesn't seem intentional

helpdesk_mgmt_stage_server_action/tests/test_helpdesk_ticket_stage.py
```diff
@@ -1,96 +1,45 @@
...
-        self.assertNotEqual(self.helpdesk_ticket_3.user_id, self.env.user)
-        self.helpdesk_ticket_2.write({"stage_id": self.helpdesk_ticket_stage_2.id})
-        self.helpdesk_ticket_3.write({"stage_id": self.helpdesk_ticket_stage_2.id})
-        self.assertEqual(self.helpdesk_ticket_2.user_id, self.env.user)
-        self.assertEqual(self.helpdesk_ticket_3.user_id, self.env.user)
-        self.helpdesk_ticket_3.write({"user_id": False})
-        self.helpdesk_ticket_3.write({"stage_id": self.helpdesk_ticket_stage_2.id})
-        self.assertFalse(self.helpdesk_ticket_3.user_id)
 
-    def test_helpdesk_ticket_without_stage(self):
```
ivantodorovich: I would very much prefer to keep the existing unit tests during a migration. It helps to ensure the behavior is not unwantedly modified.

It's ok to refactor, but the same features should be tested (unless we decide to change them of course, but I think this isn't the case). I think a feature here is lost: not triggering the action when the stage doesn't actually change, as mentioned in the previous comment

ivantodorovich (changes requested):

yankinmax:
@ivantodorovich I wasn't sure to understand the cases where we need to check the stage is changed, so I have simplified the code. Thanks for pointing to this, I'll update.

helpdesk_mgmt_stage_server_action/models/helpdesk_ticket.py
yankinmax: I've returned the check for different stages on the tickets, but the test you mentioned is checking something else and seems useless. The test I've written is almost enough to check the server action is triggered to run.

ivantodorovich (approved):

ivantodorovich:
travis is red, could you check? 🙏🏻

imlopes (approved):

ivantodorovich:
@yankinmax base PR is merged, could you remove the test dependency?
- https://github.com/OCA/helpdesk/pull/638

gurneyalex (approved):

gurneyalex:
Can you check why we have both README.md and README.rst?

marcelsavegnago (approved):

marcelsavegnago:
> Can you check why we have both README.md and README.rst?

ping @yankinmax

PR #652, opened 2024-11-05
[17.0][MIG] helpdesk_mgmt_fieldservice

imlopes:
Based on this PR: https://github.com/OCA/helpdesk/pull/608


helpdesk_mgmt_fieldservice/models/res_partner.py
```diff
@@ -10,10 +10,10 @@ def action_view_helpdesk_tickets(self):
         ) and self.env.user.has_group("helpdesk_mgmt.group_helpdesk_user_own"):
             context = dict(self.env.context)
             context.pop("search_default_open", None)
-            self = self.with_context(
+            context_self = self.with_context(
```
yankinmax, line 13: This change is useless. I would instead just clean the super call as self will be updated with context anyway:
`return super().action_view_helpdesk_tickets()`
So we need only one return statement in the end

yankinmax (approved):

PR #655, opened 2024-11-12
[17.0][MIG] helpdesk_mgmt_stage_validation: Migration to 17.0

yankinmax:


ivantodorovich:
The depended PR is merged:
- https://github.com/OCA/helpdesk/pull/638

Could you update the this PR? 🙏🏻

helpdesk_mgmt_stage_validation/models/helpdesk_ticket.py
```diff
@@ -20,9 +20,9 @@ def _check_ticket_has_empty_fields(self):
         fields = ", ".join(fields)
         if fields:
             error_message = _(
-                "Ticket %s can't be moved to the stage %s until "
-                "the following fields are set: %s."
-            ) % (self.name, self.stage_id.name, fields)
+                f"Ticket {self.name} can't be moved to the stage "
+                f"{self.stage_id.name} until the following fields are set: {fields}."
+            )
```
ivantodorovich, line 25: This is not correct. If you use an f-string, it will be substituted before calling gettext translation, and so the translation won't work.

The previous implementation was ok, could you revert it?

ivantodorovich (changes requested):

ivantodorovich:
Superseeded by:
- https://github.com/OCA/helpdesk/pull/661

PR #669, opened 2024-12-17
[16.0][OU-FIX] helpdesk_mgmt_project: Migration script was not working when migrating from a previous major version

etobella:
@victoralmau


kluna1998 (approved):
LGTM!

victoralmau (approved):

pedrobaeza (approved):
You can use also `openupgrade.column_exists`, but it's OK.

PR #681, opened 2025-01-15
[16.0][ADD] helpdesk_partner_team_category: New module helpdesk_partn…

lbarry-apsl:
This module allows you to restrict tems and categories for each coustumer.

cc https://github.com/APSL 165663

@miquelalzanillas @peluko00  @mpascuall @ppyczko @javierobcn @BernatObrador please review


mpascuall:
It would be good for the module to depend on res.partner and for the view to create a new ticket to apply a domain so that only the categories selected in the contact are displayed.

helpdesk_partner_team_category/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module allows you to restrict tems and categories for each coustumer.
```
BernatObrador, line 1: Please correct the description.
tems -> teams, coustumer -> customer

miquelalzanillas:
Hi @lbarry-apsl ,

Thank you for this. I find the module name a bit confusing, as its main functionality is related to the user portal. Additionally, I believe this module should be split into two separate modules to better reflect its purpose.

For example:

- helpdesk_portal_team_restriction

- helpdesk_portal_category_restriction

helpdesk_partner_team_category/readme/DESCRIPTION.md
lbarry-apsl: Done in #691

lbarry-apsl:
> It would be good for the module to depend on res.partner and for the view to create a new ticket to apply a domain so that only the categories selected in the contact are displayed.

Hi @mpascuall 

Now looking at the code I see that it is not necessary to depend on res.partner, because it is a model that is in the base module.

And applying the domain filter in the internal user view is not what this module intends. This module only limits the portal user, an internal user can modify without restriction. I think it is a good option to allow exceptional changes to be made.

lbarry-apsl:
> Hi @lbarry-apsl ,
> 
> Thank you for this. I find the module name a bit confusing, as its main functionality is related to the user portal. Additionally, I believe this module should be split into two separate modules to better reflect its purpose.
> 
> For example:
> 
>     * helpdesk_portal_team_restriction
> 
>     * helpdesk_portal_category_restriction

Hi @miquelalzanillas ,

I have changed the name of de module in this new PR #691 

Doing it in two separate modules is not possible because they are overwriting the controller of the helpdesk_mgmt module in the create_new_ticket() function.

If I separate it in two modules I have to make one depend on the other otherwise both cannot be used at the same time

lbarry-apsl:
Correction in #691

PR #683, opened 2025-01-15
[17.0][DCK] helpdesk_mgmt: Change development_status key to Production/Stable

victoralmau:
Change development_status key to Production/Stable

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #688, opened 2025-01-17
[17.0][FIX] helpdesk_mgmt: Correct display of avatar images in the portal

victoralmau:
Correct display of avatar images in the portal

**Before**
antes

**After**
despues

@Tecnativa


pedrobaeza (approved):
Fast-tracking it as straight-forward:

/ocabot merge patch

PR #693, opened 2025-02-04
[17.0][IMP] heldpesk*: Update logos

etobella:
Updating logs to the new format

PR #676, opened 2025-01-05
[18.0][MIG] helpdesk_mgmt: Migration to 18.0

kobros-tech:
this PR is just a fixing to PR #664 which needed to update the images, kanban card content, and kanban menu content

PR #694, opened 2025-02-05
fix commit message for #693

kobros-tech:
The commit message for #693 is:
 [17.0][IMP] heldpesk*: Update logos #693 
 
 which is syntactical incorrect and can not be shown in log using grep as a result.
the commit message should be:
[IMP] helpdesk_mgmt: Update Logo

This will be appended in the #676


pedrobaeza:
No, it's correct and it has been used a lot when you work in several modules at the same time. You should grep other way, or masking the `*`.

Closing as no valid.

kobros-tech:
> No, it's correct and it has been used a lot when you work in several modules at the same time. You should grep other way, or masking the `*`.
> 
> Closing as no valid.

@pedrobaeza 
What does "heldpesk" mean, then?

pedrobaeza:
Oh, you mean that... I though you were talking about the `*`. Well, doing a merge commit this way is not the solution. That only mangles the commit history, and we don't want that. They way to see the commits affecting a folder is to use the equivalent to https://github.com/OCA/helpdesk/commits/17.0/helpdesk_mgmt

kobros-tech:
> Oh, you mean that... I though you were talking about the `*`. Well, doing a merge commit this way is not the solution. That only mangles the commit history, and we don't want that. They way to see the commits affecting a folder is to use the equivalent to https://github.com/OCA/helpdesk/commits/17.0/helpdesk_mgmt

does it mean we should keep it like that?

kobros-tech:
commits are affecting the module directory, no problem I saw an error and wanted to fix it.

best regards 🤍

pedrobaeza:
Yeah, thanks, but that provokes other side effects. Think that a lot of people is not caring too much about the commit messages, and well, typos may happen continuously and pass unnoticed in the review. That's why I recommend you to not use that way of checking changes in a module.

kobros-tech:
@pedrobaeza 
no worries 
I guess, I am still able to fix it by getting the migration PR merged, right?

pedrobaeza:
Yeah, you can cherry-pick the commit and fix the commit message.

PR #724, opened 2025-04-16
[17.0][IMP] helpdesk_mgmt: Reposition 'New Ticket' Button

AlvaroRM11:
@BinhexTeam

This is the before and after of the button position:

Before:
before

After:
after


christian-ramos-tecnativa (approved):
IMO it looks better there, thanks

jelenapoblet (approved):
Small change, but it makes the difference, i like it.

pedrobaeza (approved):
/ocabot merge patch

PR #725, opened 2025-04-17
[17.0][IMP] helpdesk_type: Portal ticket type

juancarlosonate-tecnativa:
- [x] Add configuration option to display the type in the new ticket portal view (portal form).
- [x] Add the type field to the portal ticket list (if the previous option is enabled).
- [x] Add the field to the portal ticket detail view.

@tecnativa TT56029
@victoralmau @pedrobaeza @david-banon-tecnativa

### DIFF
--- a/helpdesk_type/models/helpdesk_ticket_type.py
+++ b/helpdesk_type/models/helpdesk_ticket_type.py
@@ -16,3 +16,9 @@ class HelpdeskType(models.Model):
         string="Teams",
         help="Helpdesk teams allowed to use this type.",
     )
+    active = fields.Boolean(default=True)
+    show_in_portal = fields.Boolean(
+        string="Show type in portal form",
+        default=True,
+        help="Allow to select type when creating a new ticket in the portal.",
+    )
--- a/helpdesk_type/models/res_company.py
+++ b/helpdesk_type/models/res_company.py
@@ -1,0 +1,13 @@
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    helpdesk_mgmt_portal_type = fields.Boolean(
+        string="Show type in portal form",
+        help="Select type in Helpdesk portal",
+    )
+    helpdesk_mgmt_portal_type_id_required = fields.Boolean(
+        string="Required Type field in Helpdesk portal",
+    )
--- a/helpdesk_type/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_type/views/helpdesk_ticket_templates.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="portal_ticket_list" inherit_id="helpdesk_mgmt.portal_ticket_list">
+        <xpath expr="//th[@t-if=&quot;groupby != 'stage'&quot;]" position="after">
+            <th
+                t-if="request.env.company.helpdesk_mgmt_portal_type"
+                name="th_ticket_type_column"
+            >Type</th>
+        </xpath>
+        <xpath expr="//td[@t-if=&quot;groupby != 'stage'&quot;]" position="after">
+            <td
+                t-if="request.env.company.helpdesk_mgmt_portal_type"
+                name="td_ticket_type_value"
+            >
+                <span t-esc="ticket.type_id.name" />
+            </td>
+        </xpath>
+    </template>
+    <template id="portal_create_ticket" inherit_id="helpdesk_mgmt.portal_create_ticket">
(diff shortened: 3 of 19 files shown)


david-banon-tecnativa (commented):
Translations are missing, but otherwise looks good to me

helpdesk_type/models/helpdesk_ticket_type.py
```diff
@@ -16,3 +16,9 @@ class HelpdeskType(models.Model):
         string="Teams",
         help="Helpdesk teams allowed to use this type.",
     )
+    active = fields.Boolean(default=True)
+    show_in_portal = fields.Boolean(
+        string="Show type in portal form",
+        default=True,
+        help="Allow to select type when creating a new ticket in the portal.",
+    )
```
victoralmau, line 24:
```suggestion
    helpdesk_mgmt_portal_select_type = fields.Boolean(
        string="Show type in portal form",
        default=True,
        help="Select type in Helpdesk portal",
    )
    helpdesk_mgmt_portal_type_id_required = fields.Boolean(
        string="Required Type field in Helpdesk portal",
        default=True,
    )
```

helpdesk_type/views/helpdesk_ticket_templates.xml
```diff
@@ -0,0 +1,47 @@
...
+                    </div>
+                </div>
+        </xpath>
+    </template>
+    <template
+        id="portal_helpdesk_ticket_page"
+        inherit_id="helpdesk_mgmt.portal_helpdesk_ticket_page"
+    >
+        <xpath expr="//span[@t-field='ticket.create_date']/.." position="after">
+            <div t-if="ticket.type_id">
```
victoralmau:
```suggestion
            <div t-if="ticket.type_id" name="ticket_type_id">
```
Add similar identifiers/name in the different sites, they are very useful in the future if you want to add something in a specific site.

victoralmau (changes requested):
If you need to, add an initial commit to `helpdesk_mgmt` to add identifiers/name to the portal elements you need to make selectors easier and more useful in the future.

helpdesk_type/models/helpdesk_ticket_type.py
juancarlosonate-tecnativa: show_in_portal is there in case, in the future, some type should not be shown on the portal, i don't think it should be removed, as it's consistent with the core practice in Helpdesk.

helpdesk_mgmt_portal_select_type is defined at the company level so that the configuration is company-dependent, i don't think it's appropriate to break the consistency of these concepts.

victoralmau: Sorry, I thought that was the name of the field indicated in `res_company`, it is correct.

helpdesk_type/models/res_company.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    helpdesk_mgmt_portal_select_type = fields.Boolean(
+        string="Show type in Helpdesk portal list and form"
+    )
```
victoralmau:
```suggestion
    helpdesk_mgmt_portal_type = fields.Boolean(
        string="Show type in portal form",
        help="Select type in Helpdesk portal",
    )
    helpdesk_mgmt_portal_type_id_required = fields.Boolean(
        string="Required Type field in Helpdesk portal",
    )
```

juancarlosonate-tecnativa:
ping @victoralmau @david-banon-tecnativa some changes

david-banon-tecnativa:
Portal user can't create ticket when show type in portal is enabled because it has no permissions to view types, missed that on my last review

juancarlosonate-tecnativa:
ping @david-banon-tecnativa @victoralmau

david-banon-tecnativa (approved):
Looks good to me

victoralmau (commented):
Functional review OK.

IMO I don't like defining the options this way
ejemplo

I think it would be more appropriate to create a "_section_" for each field (teams, types and categories). 
ejemplo2

What do you think @pedrobaeza ?

pedrobaeza:
Well, if everything is related a the "new ticket" form in portal, it's OK to go in the same config section, although being one for teams and another for types, but texts should be homogenized and texts rationalized for UX.

juancarlosonate-tecnativa:
like this? @pedrobaeza @victoralmau 
image

juancarlosonate-tecnativa:
ping @pedrobaeza review and merge please?

victoralmau (approved):

marcelsavegnago (approved):

PR #723, opened 2025-04-16
[17.0][IMP] helpdesk_mgmt: ticket categories hierarchy support

DantePereyra:
With this improvement, ticket category becomes hierarchical.

PR #728, opened 2025-04-23
[17.0][IMP] helpdesk_mgmt: Add team hierarchy

juancarlosonate-tecnativa:
Add team parent and compute the display name

@tecnativa TT56101
@david-banon-tecnativa @victoralmau

### DIFF
--- a/helpdesk_mgmt/README.rst
+++ b/helpdesk_mgmt/README.rst
@@ -226,6 +226,7 @@ Contributors
   - Víctor Martínez
   - Carolina Fernandez
   - Carlos Roca
+  - Juan Carlos Oñate
 
 - `ID42 Sistemas <https://www.id42.com.br>`__:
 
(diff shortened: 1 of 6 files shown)


david-banon-tecnativa (approved):
Looks good to me

victoralmau (approved):

pedrobaeza (changes requested):
Please include tests for the display name with hierarchy, and shouldn't you define `display_name` with `recursive=True`?

dalonsod:
@juancarlosonate-tecnativa you could take a look at https://github.com/OCA/helpdesk/pull/723 , that makes the same improvement with ticket categories (possibly with Pedro's suggestions), and review it as well

juancarlosonate-tecnativa:
> @juancarlosonate-tecnativa you could take a look at #723 , that makes the same improvement with ticket categories (possibly with Pedro's suggestions), and review it as well

Thank you very much, you're right. Also, the way you did it is just like the core does it.

juancarlosonate-tecnativa:
@pedrobaeza @victoralmau @david-banon-tecnativa @dalonsod Review again, please and i realized that category_ids is not on the form, so it wasn't possible to add a category to the team

pedrobaeza (approved):

david-banon-tecnativa (approved):

victoralmau (approved):

PR #735, opened 2025-04-28
[18.0][FIX] helpdesk_mgmt: Embed "New Ticket" portal button properly

ypapouin:
This commit creates a new action container in "portal.portal_docs_entry" and "portal.portal_searchbar". This way, the "New Ticket" button is easily added to the container without breaking the existing design layout.

**Before**
image
image

**After**
image
image
image

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
@@ -22,6 +22,28 @@
             </li>
         </xpath>
     </template>
+    <template id="portal_ticket_new_button">
+        <form method="POST" t-attf-action="/new/ticket">
+            <input type="hidden" name="csrf_token" t-att-value="request.csrf_token()" />
+            <button
+                name="create_new_ticket"
+                type="action"
+                t-att-class="'btn btn-primary ' + (_btn_classes or '')"
+            >
+                <i class="fa fa-plus" /> New</button>
+        </form>
+    </template>
+    <template
+        id="portal_docs_entry"
+        inherit_id="portal.portal_docs_entry"
+        primary="True"
+    >
+        <xpath expr="//span[@t-out='title']" position="after">
+            <t t-call="helpdesk_mgmt.portal_ticket_new_button">
+                <t t-set="_btn_classes" t-value="'btn-sm'" />
+            </t>
+        </xpath>
+    </template>
     <template
         id="portal_my_home"
         name="Portal My Home : ticket entries"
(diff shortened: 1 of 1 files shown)


AlvaroRM11 (approved):
Functionality and code quality both look good. Well done!

etobella (approved):
Functionally looks great and thanks for the fix.

Just a small comment (non-blocking but recommended) Change the action name from `action` to `helpdesk_action` in order to avoid possible collisions in the future.

helpdesk_mgmt/views/helpdesk_ticket_templates.xml
```diff
@@ -37,42 +44,55 @@
...
-                    class="btn btn-primary"
-                    style="float: right; margin-right: 0px; margin-bottom: 5px;"
-                >New Ticket</button>
-            </form>
         </div>
     </template>
+    <template id="portal_searchbar" inherit_id="portal.portal_searchbar">
+        <xpath expr="//div[@t-if='searchbar_sortings']" position="before">
+            <t t-if="action">
+                <t t-out="action" />
```
pedrobaeza: Why not putting here directly `<form>` and so on instead of having to declare a variable?

ypapouin: The purpose of this was to keep the initial logic where everything was set using `t-set`.

pedrobaeza: Well, I think it's not needed and complicate the code.

ypapouin: Ok, do I also need to edit the other one (`portal_docs_entry`) ?

pedrobaeza: I think the goal of this refactoring is to embed everything of the button in one template, and then only do the `t-call` on each place where needed, so yes.

ypapouin: I thought about moving the form in its own template and using a `t-call` to display it but there is a minor difference between both buttons (`btn-sm` class is added for the button in the "My account" display). Is there a better way to handle it ?

In fact, if I move the `<form>`, I would still need a variable to enable/disable its display, otherwise
image

pedrobaeza: You can pass by variables some of that values, like the class or the enabling, but reduce it to the minimum. Anyway, I see weird that you need to do that instead of just inheriting from the specific button and that's it.

ypapouin:
@pedrobaeza , changes are made as your request:
- Template for the "New Ticket" button
- Class variable to set the small button 
- `portal_docs_entry` and `portal_searchbar` are now `primary` to get ride of a context variable

pedrobaeza (approved):
Thanks for the efforts

/ocabot merge minor

pedrobaeza:
@AlvaroRM11 ping from @jelenapoblet :wink:

PR #729, opened 2025-04-23
[17.0][FIX] helpdesk_mgmt: Fix user_id when new ticket is created

peluko00:
According with this commit https://github.com/odoo/odoo/pull/138213 `alias_user_id` is removed in Odoo 17.

cc https://github.com/APSL 9141
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall @BernatObrador @ppyczko please review

PR #737, opened 2025-04-30
[18.0][FIX] helpdesk_mgmt: Fix user_id when new ticket is created

victoralmau:
FWP from 17.0: https://github.com/OCA/helpdesk/pull/729

According with this commit https://github.com/odoo/odoo/pull/138213 alias_user_id is removed in Odoo 17.

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #744, opened 2025-05-08
[18.0][MIG] helpdesk_mgmt_rating

tarteo:


marcelsavegnago (approved):

helpdesk_mgmt_rating/models/helpdesk_ticket.py
```diff
@@ -82,12 +82,15 @@ def rating_get_ticket_id(self):
...
+        self.ensure_one()
         action = self.env["ir.actions.act_window"]._for_xml_id(
             "helpdesk_mgmt_rating.helpdesk_ticket_rating_action"
         )
         action["name"] = _("Ticket Rating")
         action_context = safe_eval(action["context"]) if action["context"] else {}
         action_context.update(self.env.context)
         action_context.pop("group_by", None)
+        if not action_context.get("id"):
+            action_context["id"] = self.id
```
victoralmau, line 94: Why is this change necessary? In theory it should not be necessary.

victoralmau (changes requested):
IMO it is necessary to make the following changes:
- Purge the translation `.pot` files.
- Squash the administrative commits.
- Remove the migrations folder.

tarteo:
@victoralmau removed pot, rebased translation commits, removed migration dir

helpdesk_mgmt_rating/models/helpdesk_ticket.py
tarteo: Otherwise it won't work, feel free to try it yourself

victoralmau: I understand that, but I meant that this change should not be necessary because odoo core should take care of it; in any case, the important thing is that it works, so no problem.

victoralmau (approved):
Functional review OK

helpdesk_mgmt_rating/models/helpdesk_ticket.py
tarteo: Ah yes, true. I also expected it to work out of the box, but alas.

PR #746, opened 2025-05-12
[16.0][FIX] helpdesk_mgmt: company_id must always be available for check_company

NL66278:
In a previous commit check_company=True was added to the partner_id field. company_id was added as a hidden field to the form view. Unfortunately company_id was already present for users having multi-company rights. When updating modules based on helpdesk_mgmt this leads to errors. This should not happen because of the hidden field, but adding it this way prevents the error.

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_views.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_views.xml
@@ -169,15 +169,22 @@
                     <group name="main">
                         <group>
                             <field name="active" invisible="1" />
-                            <field name="company_id" invisible="1" />
                             <field name="team_id" options='{"always_reload": True}' />
                             <field name="user_ids" invisible="1" readonly="1" />
                             <field name="user_id" options='{"always_reload": True}' />
                             <field name="priority" widget="priority" />
+                            <!-- Because company_id is used for check_company
+                                 the field should always be present, even if invisible.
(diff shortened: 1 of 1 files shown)


flotho (approved):
Thanks
Easy code review, 
LGTM

marcelsavegnago (approved):

pedrobaeza (changes requested):
This change is not adding value. The field was previously invisible always, so no need to put a negative group.

flotho:
> This change is not adding value. The field was previously invisible always, so no need to put a negative group.

Thanks @pedrobaeza , but initially, the field will be present twice in the view when you're in multi company group ? Am I wrong

pedrobaeza:
Yes, and that's not any problem. Since a lot of time (I would say 10.0), there's official support for having the same field the times you want in the same view.

NL66278:
@pedrobaeza The fact is that with this change I can update a dependent module, while without the change I get an error. As I already said in the PR description "this should not happen". Unfortunately, it does...

pedrobaeza:
The error may be another thing due to selectors orders. If you don't put here the error, we can't help you more. As it is now, as said, this can't be accepted.

NL66278:
I cannot reproduce the original issue any more. Might have been a peculiarity in Odoo that has been resolved in the meantime. Closing this.

PR #752, opened 2025-05-28
[18.0][IMP] helpdesk_mgmt: Add team hierarchy

juancarlosonate-tecnativa:
FWP from 17.0: https://github.com/OCA/helpdesk/pull/728

@tecnativa TT56101

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket_team.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket_team.py
@@ -7,6 +7,10 @@ class HelpdeskTeam(models.Model):
     _description = "Helpdesk Ticket Team"
     _inherit = ["mail.thread", "mail.alias.mixin"]
     _order = "sequence, id"
+    _parent_name = "parent_id"
+    _parent_store = True
+    _parent_order = "name"
+    _rec_name = "complete_name"
 
     sequence = fields.Integer(default=10)
     name = fields.Char(required=True)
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -64,6 +68,23 @@ class HelpdeskTeam(models.Model):
         default=True,
         help="Allow to select this team when creating a new ticket in the portal.",
     )
+    parent_id = fields.Many2one(
+        "helpdesk.ticket.team", string="Parent Team", index=True
+    )
+    complete_name = fields.Char(
+        compute="_compute_complete_name", store=True, recursive=True
+    )
+    parent_path = fields.Char(index=True)
```
christian-ramos-tecnativa, line 77: It seems like this field is not used
```suggestion
```

christian-ramos-tecnativa (changes requested):

helpdesk_mgmt/models/helpdesk_ticket_team.py
juancarlosonate-tecnativa: If i'm not wrong, parent_path is required because of _parent_store = True, it's the field where the ids are stored with their hierarchy

victoralmau (approved):

helpdesk_mgmt/models/helpdesk_ticket_team.py
pedrobaeza: Yes!

ivantodorovich (approved):

marcelsavegnago (approved):

PR #761, opened 2025-06-11
[16.0][FIX] helpdesk_mgmt: Use t-field="ticket.partner_id" to avoid access errors in portal

victoralmau:
Use `t-field="ticket.partner_id"` to avoid access errors in portal

Example use case:
- Create a ticket to partner Marc Demo
- Add as follower a portal user (Joel Wills)
- Login to the portal with the user Joel Wills and go to the ticket list page
- No error will be displayed and the partner's name will be displayed.

@Tecnativa TT56075

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
@@ -117,7 +117,7 @@
                         <t t-foreach="tickets" t-as="ticket">
                             <tr>
                                 <td class="text-left">
-                                    <span t-esc="ticket.partner_id.name" />
+                                    <span t-field="ticket.partner_id" />
                                 </td>
                                 <td class="text-left">
                                     <span t-esc="ticket.number" />


pedrobaeza (commented):
The problem with this is that the full display name will be shown in case the contact is a children one, changing behavior. An alternative is to put `sudo` on the `t-esc` expression.

victoralmau:
> The problem with this is that the full display name will be shown in case the contact is a children one, changing behavior. An alternative is to put `sudo` on the `t-esc` expression.

Ok, changed, although in the ticket page the full name is shown (even if the widget is used). https://github.com/OCA/helpdesk/blob/a39fa9fe9f09b8e73899d42fd8d2dc15c5df5f66/helpdesk_mgmt/views/helpdesk_ticket_templates.xml#L321

pedrobaeza:
If that's the case, then let's homogenize to one or the other. You choose.

victoralmau:
> If that's the case, then let's homogenize to one or the other. You choose.

Ok, my initial approach then, to be consistent with the data currently displayed on the ticket page.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #766, opened 2025-06-18
[17.0][ADD] helpdesk_timesheet_date_time_order: New module helpdesk_t…

lbarry-apsl:
Add new to the top of the list, rather than the bottom, and sort by date time
in descending order.

cc https://github.com/APSL #HT02143

@miquelalzanillas @mpascuall @javierobcn @peluko00 @BernatObrador @ppyczko please review


dalonsod:
Hello @lbarry-apsl this could be a good improvement for base `helpdesk_mgmt_timesheet` addon IMO.

Obviously, this functionality would slightly change current behavior, that is a good point for an `ADD` instead an `IMP`. At the other side, it's a very small change, that could be adopted without any problem for current `helpdesk_mgmt_timesheet` users and maintainers, what do you think?

lbarry-apsl:
Hi @dalonsod ,

Very good suggestion, but I think some users prefer the original order. Don't you think so?

dalonsod:
Well, as one of the original authors I'd directly approve such thing as an `IMP`, but I'm not the only one... we could ask another usual developers. Anyway, feel free to go ahead with this addon.

lbarry-apsl:
Okay, so I'm going to open another PR with these changes to the module `helpdesk_mgmt_timesheet`as an improvement, to see if there's acceptance.

PR #770, opened 2025-06-19
[17.0][IMP] helpdesk_mgmt_timesheet: Editable top and default order 'project_timesheet_time_control'

lbarry-apsl:
Editable top and default order 'project_timesheet_time_control'

cc https://helpdesk.apsl.net/ HT02143

@dalonsod @miquelalzanillas @peluko00 @mpascuall @ppyczko @javierobcn @BernatObrador please review

There is another option which is to keep it in a separate module: #766

### DIFF
--- a/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
+++ b/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
@@ -136,7 +136,7 @@
                         name="timesheet_ids"
                         context="{'default_project_id': project_id, 'default_task_id': task_id}"
                     >
-                        <tree editable="bottom" delete="true" default_order="date_time">
+                        <tree editable="top" delete="true">
                             <field name="company_id" invisible="1" />
                             <field name="project_id" invisible="1" />
                             <field name="task_id" invisible="1" />


mpascuall (approved):
LGTM!

dalonsod (changes requested):
Only cosmetic change: fix typos for commit title in "Edditable" and "defautl", and for PR title as well

dalonsod (approved):
:+1:

ppyczko (approved):
LGTM!

BernatObrador (approved):
LGTM

pedrobaeza (changes requested):
The default order shouldn't be removed, as if not, they won't appear properly ordered.

lbarry-apsl:
> The default order shouldn't be removed, as if not, they won't appear properly ordered.

@pedrobaeza , The module depends on the project_timesheet_time_control module which already does the sorting in the model: https://github.com/OCA/project/blob/17.0/project_timesheet_time_control/models/account_analytic_line.py#L16

pedrobaeza:
But it's not the same order, as the other is desc, and here you are also modifying that order, and that's why you are putting also "top", so I don't fully agree with this change, as it's not aligned with what core does with task timesheets, and this one is mimicking that one.

lbarry-apsl:
> But it's not the same order, as the other is desc, and here you are also modifying that order, and that's why you are putting also "top", so I don't fully agree with this change, as it's not aligned with what core does with task timesheets, and this one is mimicking that one.

@pedrobaeza , How do you see a separate module option in this PR #766 ?

My requirement it is sorted by descending date and time, without the changes it is sorted by ascending date.

pedrobaeza:
In this case I'm afraid you should do an extra glue module for this part, but it's not OK to add the configuration option here for complicating while the standard is not doing that.

PR #771, opened 2025-06-20
[18.0] [FIX] helpdesk_motive add permission for user_own and user_team groups

andreampiovesana:
security for user_own and user_team groups otherwise user with different permissions are not able to set motive in a ticket


dalonsod:
Hello @andreampiovesana are really needed `create` and `write` permissions? And setting permissions for `group_helpdesk_user_team` is not required, as it actually inherits from `group_helpdesk_user_own`.

andreampiovesana:
change as you want, but now a user_team can't set a motive

dalonsod:
Tested in runboat, and it works even if ACL for team group is removed, as expected. Please note that motives depend on team, you may tested sometimes creating a ticket selecting a team that could not have available motives. The only change that should be set is moving current ACL team for user group to user_own group (line 2 at .csv file).

And please remember the other question: why do you add create and write permissions? Only setting read permissions should be enough IMO

andreampiovesana:
merge?

PR #772, opened 2025-06-25
[17.0][IMP] helpdesk_mgmt: Fine-tuning of the Kanban view for helpdesk.ticket

carlos-lopez-tecnativa:
- Use the `many2one_avatar_user` widget to distinguish between users with profile images and those without. This widget displays a colored circle with the user's initial instead of showing a blank image when no profile picture is set.

- Add the category field to the Kanban view.

TT56895

image

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_views.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_views.xml
@@ -171,7 +171,11 @@
                             <field name="active" invisible="1" />
                             <field name="team_id" options='{"always_reload": True}' />
                             <field name="user_ids" invisible="1" readonly="1" />
-                            <field name="user_id" options='{"always_reload": True}' />
+                            <field
+                                name="user_id"
+                                options='{"always_reload": True}'
+                                widget="many2one_avatar_user"
+                            />
                             <field name="priority" widget="priority" />
                             <field
                                 name="company_id"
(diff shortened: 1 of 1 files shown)


helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -315,6 +322,12 @@
                             <div class="o_kanban_record_top">
                                 <div class="o_kanban_record_headings">
                                     <field name="name" />
+                                    <span
+                                        t-if="record.category_id.value"
+                                        style="display: block; margin-top: 4px;"
+                                    >
+                                        <field name="category_id" />
```
pedrobaeza, line 330: Better to put it on "disabled" color for being differentiated, like some fields in tasks:

imagen

helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -272,6 +276,9 @@
                 <field name="name" />
                 <field name="partner_name" />
                 <field name="user_id" />
+                <field name="user_ids" />
+                <field name="team_id" />
```
victoralmau, line 280:
```suggestion
```
Just a small detail, is it necessary to add these fields?

victoralmau (approved):
Functional review OK

helpdesk_mgmt/views/helpdesk_ticket_views.xml
carlos-lopez-tecnativa: Yes, they are necessary. Otherwise, if you try to assign a user from the kanban view, you get an error because this field has a dynamic domain.
image

image

helpdesk_mgmt/views/helpdesk_ticket_views.xml
carlos-lopez-tecnativa: Done

pedrobaeza (approved):
/ocabot merge minor

PR #776, opened 2025-06-27
[17.0][IMP] helpdesk_mgmt: Improve user and team assignment in helpdesk tickets

carlos-lopez-tecnativa:
- Assign the current user by default if no user is set.
- Assign the first team of the user.
- When the team is changed, if the user is not in the team, unset the user.

TT56910

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -25,8 +25,18 @@ class HelpdeskTicket(models.Model):
     @api.depends("team_id")
     def _compute_user_id(self):
         for ticket in self:
-            if not ticket.user_id and ticket.team_id:
-                ticket.user_id = ticket.team_id.create_uid
+            if not ticket.user_id:
+                ticket.user_id = self.env.user
+            if ticket.team_id and ticket.user_id not in ticket.team_id.user_ids:
+                # If the user is not part of the team, we remove the user
+                ticket.user_id = False
+
+    @api.depends("user_id")
+    def _compute_team_id(self):
+        for ticket in self:
+            if not ticket.team_id and ticket.user_id.helpdesk_team_ids:
+                # If no team is set, we default to the user's first team
+                ticket.team_id = ticket.user_id.helpdesk_team_ids[0]
 
     @api.model
     def _read_group_stage_ids(self, stages, domain, order):
@@ -52,6 +62,9 @@ class HelpdeskTicket(models.Model):
         string="Assigned user",
         tracking=True,
         index=True,
+        compute="_compute_user_id",
+        store=True,
+        readonly=False,
         domain="team_id and [('share', '=', False),('id', 'in', user_ids)] or [('share', '=', False)]",  # noqa: B950,E501
     )
     user_ids = fields.Many2many(
@@ -104,6 +117,9 @@ class HelpdeskTicket(models.Model):
         comodel_name="helpdesk.ticket.team",
         string="Team",
         index=True,
+        compute="_compute_team_id",
+        store=True,
+        readonly=False,
     )
     priority = fields.Selection(
         selection=[
--- a/helpdesk_mgmt/tests/test_helpdesk_ticket.py
+++ b/helpdesk_mgmt/tests/test_helpdesk_ticket.py
@@ -1,5 +1,7 @@
 import time
 
+from odoo.tests.common import Form
+
 from .common import TestHelpdeskTicketBase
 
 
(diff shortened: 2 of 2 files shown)


helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -22,11 +22,25 @@ def _compute_stage_id(self):
...
+            ):
+                # If no team is set, we default to the user's first team
+                ticket.team_id = ticket.user_id.helpdesk_team_ids[0]
+            if (
+                ticket.user_id
+                and ticket.team_id
+                and ticket.user_id not in ticket.team_id.user_ids
+            ):
+                # If the user is not part of the team, we remove the user
+                ticket.user_id = False
```
victoralmau:
```suggestion
new_ticket_user = self.env["res.users"]
ticket_user = ticket.user_id
            if not ticket.user_id:
                ticket_user = new_ticket_user = self.env.user
            if (
                not ticket.team_id
                and ticket_user
                and ticket_user.helpdesk_team_ids
            ):
                # If no team is set, we default to the user's first team
                ticket.team_id = ticket_user.helpdesk_team_ids[0]
            if (
                new_ticket_user
                and ticket.team_id
                and new_ticket_user in ticket.team_id.user_ids
            ):
                ticket.user_id = new_ticket_user
```

victoralmau: IMO it is not a good idea to assign something and then “remove” it (also assigning False can have side effects in some cases).

carlos-lopez-tecnativa: Your suggestion does not produce the expected behavior.
When you create a new ticket, it has no `team_id`, so the code from line 39 is not applicable and the user is never set.
With my code, this works as expected.

**With your suggestion:**
image

**With the current code**, even if the current user has no team, the current user is still assigned:
image

Regarding setting the user to False, try to handle this case:

- Create a new ticket, the current user is automatically assigned.
- Change the team to another one where the current user is not a member.

In this case, I think keeping the current user is wrong because they are no longer part of the team members.
For that reason, the user is reset to `False`. Keep in mind that the user_id field has a domain that only allows selecting users who are members of the team.
If we keep a user who no longer belongs to this team, it introduces inconsistency, in my opinion.
What do you think?

carlos-lopez-tecnativa: @victoralmau please update your review

victoralmau: My suggestion was a rough one; I think it could be adjusted slightly to cover different cases (e.g., assigning it only if you don't have a team, or if you do have a team and the user belongs to it). My suggestion was about trying to avoid assigning something and then deleting it later; I think that could be problematic.

carlos-lopez-tecnativa: I’ve updated the code. Please let me know if it’s OK for you now.

victoralmau (approved):

marcelsavegnago (approved):

marcelsavegnago:
Just one question regarding the change proposed in this PR: is it really necessary to rename the method to _compute_user_and_team_id?

carlos-lopez-tecnativa:
> Just one question regarding the change proposed in this PR: is it really necessary to rename the method to _compute_user_and_team_id?

Hi @marcelsavegnago, it's not really necessary, but keep in mind that before this commit, the method wasn't used, the field wasn't computed, so it shouldn't have any current impact.

The reason for renaming the method is to reflect its new behavior: now it computes two fields, so the new name is more descriptive of what it does.

pedrobaeza (changes requested):
Each compute has to be independent, not joined in the same compute method.

marcelsavegnago (approved):

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -25,8 +25,19 @@ def _compute_stage_id(self):
     @api.depends("team_id")
     def _compute_user_id(self):
         for ticket in self:
-            if not ticket.user_id and ticket.team_id:
-                ticket.user_id = ticket.team_id.create_uid
+            ticket_user = ticket.user_id or self.env.user
+            if ticket.team_id and ticket_user not in ticket.team_id.user_ids:
+                # If the user is not part of the team, we remove the user
+                ticket_user = self.env["res.users"]
```
pedrobaeza:
```suggestion
                ticket_user = False
```

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -25,8 +25,19 @@ def _compute_stage_id(self):
...
+            ticket.user_id = ticket_user
```
pedrobaeza: Why we reassign always a user?

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -25,8 +25,19 @@ def _compute_stage_id(self):
...
+
+    @api.depends("user_id")
+    def _compute_team_id(self):
+        for ticket in self:
+            ticket_user = ticket.user_id or self.env.user
```
pedrobaeza: We should only do it if a user is selected.

helpdesk_mgmt/models/helpdesk_ticket.py
carlos-lopez-tecnativa: Done, please update your review

pedrobaeza (approved):
/ocabot merge patch

PR #779, opened 2025-07-03
[17.0][IMP] helpdesk_type: Make type name translatable

victoralmau:
Make type name translatable

Please @pedrobaeza and @juancarlosonate-tecnativa can you review it?

@Tecnativa

### DIFF
--- a/helpdesk_type/models/helpdesk_ticket_type.py
+++ b/helpdesk_type/models/helpdesk_ticket_type.py
@@ -10,7 +10,7 @@ class HelpdeskType(models.Model):
     _description = "Helpdesk Ticket Type"
     _order = "name asc"
 
-    name = fields.Char(required=True)
+    name = fields.Char(required=True, translate=True)
     team_ids = fields.Many2many(
         "helpdesk.ticket.team",
         string="Teams",


pedrobaeza:
A migration script is needed.

victoralmau:
> A migration script is needed.

Sorry, but what script is necessary in this case? I haven't seen any similar examples. I've also tested making this change locally, and the column in the database is converted correctly, probably because https://github.com/odoo/odoo/blob/55c7feaae93651ac6ba5a000ac2493438d2317b4/odoo/fields.py#L1709

pedrobaeza:
You have to convert the field to JSON. Check https://github.com/OCA/OpenUpgrade/blob/381caca6aac60da92bea137d0f98b39984c43dcc/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py#L115 (or maybe you can call the mentioned method directly).

victoralmau:
Sorry, but I still think it's not necessary (odoo does it by setting the translate=true field), for example because the ir_translation table doesn't exist.

pedrobaeza (approved):
OK, you're right.

marcelsavegnago (approved):

pedrobaeza (approved):
/ocabot merge patch

victoralmau:
Can you make the changes in v18 as well?

pedrobaeza:
What do you mean, @victoralmau ?

victoralmau:
Oops, sorry, I got the wrong place.

PR #780, opened 2025-07-04
[17.0][FIX] helpdesk_mgmt_timesheet: Define the show_time_control field with column_invisible=1

victoralmau:
Define the `show_time_control` field with column_invisible=1

@Tecnativa

PR #781, opened 2025-07-04
[18.0][FIX] helpdesk_mgmt_timesheet: Remove show_time_control field from the views

victoralmau:
FWP from 17.0: https://github.com/OCA/helpdesk/pull/780

Remove `show_time_control` field from the views

@Tecnativa


pedrobaeza (approved):
Since the [viewtiverse Odoo PR](https://github.com/odoo/odoo/pull/137031), we don't need to add such fields to the view indeed.

/ocabot merge patch

PR #783, opened 2025-07-10
[17.0][FIX] helpdesk_mgmt_timesheet: Incorrect filter definition

pedrobaeza:
With the filter "With activity today" in the helpdesk tickets view, you get the following error:

```
Error: Failed to evaluate the domain: [("last_timesheet_activity", "=", datetime.datetime.today())].
    fnValue is not a function
```

Because the definition is incorrect.

Now it has a correct definition.

@Tecnativa


marcelsavegnago (approved):
LGTM

victoralmau (approved):

PR #794, opened 2025-07-16
[17.0][ADD] helpdesk_mgmt_tickets_menu

DantePereyra:
This addons splits the Helpdesk ticket menu into two submenus:
- "My Tickets": Displays tickets assigned to the current user.
image

- "All Tickets": Displays all tickets.
image


pedrobaeza:
This can be done in the main module instead of one extra module.

DantePereyra:
> This can be done in the main module instead of one extra module.

Following @pedrobaeza  suggestion, an `IMP` was implemented here #796.

PR #792, opened 2025-07-15
[16.0][IMP] helpdesk_mgmt_merge: Send messages and activities. Improve links

etobella:
Improve the message to include a link and send to the new ticket all messages inside. That might have sense in the case where the ticket is created automatically by an email (for example)

PR #795, opened 2025-07-17
[18.0][IMP] helpdesk_mgmt_merge: Send messages and activities. Improve links

luisDIXMIT:
Forwardport of #792

### DIFF
--- a/helpdesk_mgmt_merge/wizard/helpdesk_ticket_merge.py
+++ b/helpdesk_mgmt_merge/wizard/helpdesk_ticket_merge.py
@@ -1,3 +1,5 @@
+from markupsafe import Markup
+
 from odoo import Command, api, fields, models
 
 
@@ -69,9 +71,19 @@ class HelpdeskTicketMerge(models.TransientModel):
         merged_tickets = self.ticket_ids - self.dst_ticket_id
         self._merge_followers(merged_tickets)
         for ticket in merged_tickets:
-            self._add_message("to", self.dst_ticket_id.number, ticket)
-        ticket_numbers = ", ".join(merged_tickets.mapped("number"))
-        self._add_message("from", ticket_numbers, self.dst_ticket_id)
+            self._add_message(
+                self.env._("to"),
+                self.dst_ticket_id._get_html_link(self.dst_ticket_id.number),
+                ticket,
+            )
+        ticket_numbers = ", ".join(
+            ticket._get_html_link(ticket.number) for ticket in merged_tickets
+        )
+        self._add_message(self.env._("from"), ticket_numbers, self.dst_ticket_id)
+        merged_tickets.message_ids.filtered(
+            lambda r: r.message_type != "notification"
+        ).write({"res_id": self.dst_ticket_id.id})
+        merged_tickets.activity_ids.write({"res_id": self.dst_ticket_id.id})
         merged_tickets.write({"active": False})
 
         return {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
See pre-commit.

pedrobaeza (commented):
Shouldn't this need `Markup` for HTML content?

luisDIXMIT:
> Shouldn't this need `Markup` for HTML content?

I'm not sure, I just forward-ported the exact commit content. Maybe @etobella can clarify better what's needed. WDYT, Enric?

luisDIXMIT:
You're right @pedrobaeza . I should have tested it before pushing. With Markup, the link is displayed correctly.

etobella (approved):

helpdesk_mgmt_merge/wizard/helpdesk_ticket_merge.py
```diff
@@ -69,9 +71,21 @@ def merge_tickets(self):
         merged_tickets = self.ticket_ids - self.dst_ticket_id
         self._merge_followers(merged_tickets)
         for ticket in merged_tickets:
-            self._add_message("to", self.dst_ticket_id.number, ticket)
-        ticket_numbers = ", ".join(merged_tickets.mapped("number"))
+            self._add_message(
+                "to",
+                f"""<a href="/web#id={self.dst_ticket_id.id}&model=helpdesk.ticket">
+                {self.dst_ticket_id.number}</a>""",  # noqa: B950
```
pedrobaeza: If not, you are introducing the spaces and the line break in the string:

```suggestion
                f"<a href="/web#id={self.dst_ticket_id.id}&model=helpdesk.ticket">"
                f"{self.dst_ticket_id.number}</a>",  # noqa: B950
```

Anyway, there's a method for getting this without effort:

```suggestion
                self.dst_ticket_id._get_html_link(self.dst_ticket_id.number),
```

and the same for the rest.

luisDIXMIT: Oh, I had no idea the method existed. I used it because it’s cleaner. Thanks!

helpdesk_mgmt_merge/wizard/helpdesk_ticket_merge.py
```diff
@@ -128,7 +140,7 @@ def _add_message(self, way, ticket_numbers, ticket):
         """
         subject = "Merge helpdesk ticket"
         body = self.env._(
```
pedrobaeza: You have to split the translatable content inside `self.env._` from the parametric content. You can't put f-strings inside translate method call.

luisDIXMIT: Hi @pedrobaeza, I see your point of view. Since that line of code isn't part of my forwardport commit, should I make the change in this commit, or should I create a separate FIX commit in different PRs to port to different versions?

pedrobaeza: You can do it here, as you are touching the `Markup` thing, and check that no other one has the same problem (and thus, not being translated at all).

luisDIXMIT: Done! I tested translations. Thanks!

helpdesk_mgmt_merge/wizard/helpdesk_ticket_merge.py
```diff
@@ -69,9 +71,19 @@ def merge_tickets(self):
         merged_tickets = self.ticket_ids - self.dst_ticket_id
         self._merge_followers(merged_tickets)
         for ticket in merged_tickets:
-            self._add_message("to", self.dst_ticket_id.number, ticket)
-        ticket_numbers = ", ".join(merged_tickets.mapped("number"))
+            self._add_message(
+                "to",
```
etobella: shouldn't the to be translated too?

etobella: It might happen in other places too...

luisDIXMIT: I hope no text escapes the translations.

pedrobaeza (approved):
/ocabot merge patch

PR #797, opened 2025-07-18
[17.0][IMP] helpdesk_mgmt: allowed contacts by team

DantePereyra:
This feature allows dynamic control over which contact (partner_id) on ticket, based on the configuration of the assigned Helpdesk Team (team_id). The logic works as follows:
- If the team has a default contact, it will fill the contact field with that value.
- If the team has a unique contact, only that contact will be available.
- If the team has a list of allowed contacts, only those contacts will be available.
Team_allowed_partners
Ticket_allowed_contact

- If the team has no restrictions, all contacts will be available.
Team_without_allowed_partners
Ticket_all_contacts


rrebollo:
@DantePereyra I think I would propose this as a **standalone module**, not as an improvement to be merged into the **base workflow** — it aligns better with the **OCA mindset**.

From a **design perspective**, if I understand correctly, it overlaps with a more **transversal need**: limiting partner visibility. So I wonder if some configuration of [Sales Teams Security](https://odoo-community.org/shop/sales-teams-security-4198#attr=939347) might work even better since this restriction applies **only within the helpdesk team’s operational context** and not at the **ERP-wide level**

Technically, it looks **well implemented**.

DantePereyra:
Hi @rrebollo! I will test the addon you said and see how I can improve mine, and make a proposition as an standalone addon. Thanks for reviewing!

DantePereyra:
This PR is being closed because further tracking will take place in #878, and it changes from IMP to ADD.

PR #796, opened 2025-07-17
[17.0][IMP] helpdesk_mgmt: splits the Helpdesk ticket menu into two submenus.

DantePereyra:
By default, the Helpdesk Ticket menu displays all tickets a user can see. With this `IMP`, the menu is split into two submenus for better segmentation and usability.
- "My Tickets": Displays tickets assigned to the current user.
image
- "All Tickets": Displays all tickets.
image

PR #798, opened 2025-07-18
[18.0][IMP] helpdesk_mgmt: splits the Helpdesk ticket menu into two submenus.

DantePereyra:
FW Port of #796 

By default, the Helpdesk Ticket menu displays all tickets a user can see. With this IMP, the menu is split into two submenus for better segmentation and usability.
"My Tickets": Displays tickets assigned to the current user.
My_tickets
"All Tickets": Displays all tickets. 
All_tickets


CristianoMafraJunior (approved):

helpdesk_mgmt/views/helpdesk_ticket_menu.xml
```diff
@@ -13,6 +13,12 @@
         <field name="res_model">helpdesk.ticket</field>
         <field name="view_mode">list,kanban,form,pivot</field>
     </record>
+    <record id="helpdesk_my_tickets_action" model="ir.actions.act_window">
+        <field name="name">My Tickets</field>
+        <field name="res_model">helpdesk.ticket</field>
+        <field name="view_mode">kanban,tree,pivot,form</field>
```
dalonsod: Fix typo:
```suggestion
        <field name="view_mode">kanban,list,pivot,form</field>
```

dalonsod (approved):
:+1:

helpdesk_mgmt/views/helpdesk_ticket_menu.xml
```diff
@@ -68,9 +74,22 @@
         id="helpdesk_ticket_menu"
         name="Tickets"
         parent="helpdesk_ticket_main_menu"
-        action="helpdesk_ticket_action"
```
pedrobaeza, line 71: Can you please check if after this update, the action is reset to False removing this, or should we specify `action=""` explicitly?

DantePereyra: After verifying the configuration, the menu item remains linked to the action. Despite the fact that the action is no longer executed, his residual linkage is incorrect. Therefore, removing the action reference via a post-migration script appears to be the most appropriate solution. What do you think?

pedrobaeza: Putting `action=""` doesn't work? If not, then indeed a migration script is the way.

DantePereyra:
Migration script added, could you check again @dalonsod @CristianoMafraJunior @pedrobaeza ? Thanks!

pedrobaeza (approved):
/ocabot merge nobump

PR #800, opened 2025-07-23
[18.0][MIG] helpdesk_mgmt_sla

BhaveshHeliconia:

PR #757, opened 2025-06-02
[18.0] [MIG] helpdesk_type

mpascuall:
Module migrated to version 18.0

cc https://github.com/APSL 

@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review

PR #801, opened 2025-07-23
[18.0][MIG] helpdesk_type_sla

BhaveshHeliconia:
Depends on
- https://github.com/OCA/helpdesk/pull/800


imlopes:
depends on https://github.com/OCA/helpdesk/pull/757 too
(included on test-requirements but just to be sure that will receive reviews)

imlopes (approved):

BhaveshHeliconia:
@marcelsavegnago, could you please review this module?

marcelsavegnago:
@pedrobaeza what is the rule for accepting someone as a contributor? For example, if a user submits a standard migration PR, does that entitle them to be included in the contributors section?

pedrobaeza:
This is a bit subjective, as it can't be measured. The question is if the migration effort is significant enough or it has been only changing the version in the manifest. That's decided anyway by the PSC/maintainers of the module.

marcelsavegnago (approved):

marcelsavegnago:
> This is a bit subjective, as it can't be measured. The question is if the migration effort is significant enough or it has been only changing the version in the manifest. That's decided anyway by the PSC/maintainers of the module.

Got it, thanks for clarifying!

PR #808, opened 2025-07-30
[17.0] [IMP] Set default assigned user by category for tickets created from portal

lperez-solvos:
In the portal view, automatically assign the user specified in the Helpdesk configuration by default, but only when the ticket is created through the portal by a portal user. The list of assignable users must be limited to internal users only.
imagen
imagen
imagen
imagen

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -205,6 +205,16 @@ class HelpdeskTicket(models.Model):
                 )
                 if channel_email_id:
                     vals["channel_id"] = channel_email_id.id
+            if self.env.user.share and vals.get("category_id"):
+                category = self.env["helpdesk.ticket.category"].browse(
+                    vals["category_id"]
+                )
+                if (
+                    category.default_partner_id
+                    and not self.env.company.helpdesk_mgmt_portal_select_team
+                ):
+                    vals["user_id"] = category.default_partner_id.id
+
         return super().create(vals_list)
 
     def copy(self, default=None):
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Which is the use case for this? It seems very specific/custom.

dalonsod:
@lperez-solvos change PR & commit title, replacing "add default partner on portal" by "set default assigned user by category for tickets created from portal", that is the proper meaning of the improvement.

dalonsod:
> Which is the use case for this? It seems very specific/custom.

@pedrobaeza we think this improvement could be useful. When a customer users creates a ticket from portal frontend, they set a default category for the ticket, as their first decision of the meaning of the issue. With this new feature, it's possible to directly assign such tickets to the proper internal helpdesk user, for the very first attention (e.g. a category is specific enough and there's always the same user that should make the triage for these category tickets).

Anyway, if this feature is not interesting for a certain helpdesk installation, leaving this field empty for every category will keep current standard behavior.

pedrobaeza:
IMO, that's a usage a bit deformed according your flows, and can enter in conflict with others like teams assignation/notifications, but let's see what others think.

DantePereyra:
> IMO, that's a usage a bit deformed according your flows, and can enter in conflict with others like teams assignation/notifications, but let's see what others think.

I was testing this on Runboat, and there could be an issue when selecting a category with a user who doesn't belong to the selected team. One possible solution could be to add a restriction: if  `helpdesk_mgmt_portal_select_team` is checked, then don't assign the default user unless they are part of the selected team.
What do you think?

rrebollo:
I'm on @pedrobaeza side from a design perspective — this is a very specific case.

Assigning the responsible user in Odoo EE typically happens **at the Helpdesk Team level** as part of the system's natural flow. Configuring the **team** instead of an individual would be less intrusive and feel less like an external intervention in the process.

Your requirement is **legitimate**, but it should be addressed in the proper context.

Keeping the PR open to gather feedback from the community is reasonable.

dalonsod:
Thanks @pedrobaeza @rrebollo for the feedback. Attending your comments, we've finally moved this to a new candidate addon

* https://github.com/OCA/helpdesk/pull/868

PR #811, opened 2025-08-11
[17.0][FIX] helpdesk_mgmt: stage_id resets when you change users within the same team.

adasatorres:
Good morning, I have noticed a small issue that I believe should not be happening. When I changed the user of a ticket, its status would reset to the first stage_id available.

After investigating, I found that this seems to be related to how computed fields with dependencies work — although I might be mistaken. The compute method for the team_id field, even if it does not change the value, triggers a recomputation of the stage_id field because stage_id’s compute depends on team_id. When team_id is recalculated, Odoo detects that stage_id should also be added to the recompute queue.

A possible solution would be to change the compute method of team_id to an inverse of user_id, which would preserve the functionality while preventing this issue from occurring.

### DIFF
--- a/helpdesk_mgmt/tests/test_helpdesk_ticket.py
+++ b/helpdesk_mgmt/tests/test_helpdesk_ticket.py
@@ -1,5 +1,6 @@
 import time
 
+from odoo.tests import new_test_user
 from odoo.tests.common import Form
 
 from .common import TestHelpdeskTicketBase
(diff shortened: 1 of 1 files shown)


sheragar (approved):
LGTM. Funtional Review 👍

christian-ramos-tecnativa (approved):
Code and functional review OK

jorgeglez1990 (approved):
Funtional Review. LGTM.

pedrobaeza (changes requested):
Please don't change the field to regular field. It should stay computed writable. Also using `inverse` methods as write triggers can produce side effects, as you don't control the application order.

christian-ramos-tecnativa:
Hi @pedrobaeza the compute in the `team_id` field triggers the `_compute_stage_id` always even if the `team_id` is not changed and there is no way to control in that function if the change require a stage reset. Intead of using inverse it could be added to the write method. But removing the compute method looks mandatory in order to solve the issue

pedrobaeza:
The other compute it's not triggered if you don't assign any value, so you can instead just compare current value with the one you want to set, and if not change, don't do the set for avoiding the triggering. Something like:

```
if record.field != new_value:
    record.field = new_value
```

christian-ramos-tecnativa:
I think this may require a more deeply research, after some debugging it seems like the stage compute is getting trigger first when the user is changed. This may have something to do with the `mail.tracking.duration.mixin` assigned to the `stage_id` but I'm not sure

pedrobaeza:
Can you develop a unit test where the problem is reproducible before any patch? With it, I will check.

adasatorres:
Hi @pedrobaeza, I have created a test that reproduces the error within the addon.

Here’s a screenshot showing the error that the test produces.
Captura desde 2025-08-13 11-23-23

pedrobaeza:
"A grandes males, grandes remedios" :wink: #814

@adasatorres some advices for better tests:

- Reuse existing data if exists (in this case, we have enough with provided in the base class in common.py).
- Don't do asserts on the thing you are not testing (example: if you assign the "in progress" stage in the form, there's no need to check that this stage has been assigned. This is ORM, not the business logic you want to test. Even if there's other "wild" compute/onchange that changes that stage you have assigned, the final assert will show the possible problem. If not, you are making tests far less readable and difficult to follow what do you want to check.
- This is subjective, but on my mental scheme, when using `assertEqual`, the first argument should be the thing to check, and the second one the hardcoded expected one.
- Another stylistic subjective one (and can be general): don't leave empty lines inside a method. They increase the vertical scroll with no value. If you want to make blocks, do them with a line with a comment, not with an empty line. But most of the times, Python indentation + highlighting is enough to distinguish easily the code.

PR #812, opened 2025-08-12
[18.0] [IMP] `helpdesk_mgmt`: add action url paths

ivantodorovich:
These form nicer URLs in backend actions.

The `helpdesk-mgmt-` prefix is not ideal, but paths must be unique and so this is to avoid conflicts with Odoo's `helpdesk` module

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_menu.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_menu.xml
@@ -6,54 +6,63 @@
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">helpdesk.ticket.team</field>
         <field name="view_mode">kanban,list,form,pivot</field>
+        <field name="path">helpdesk-dashboard</field>
     </record>
     <record id="helpdesk_ticket_action" model="ir.actions.act_window">
         <field name="name">Tickets</field>
         <field name="type">ir.actions.act_window</field>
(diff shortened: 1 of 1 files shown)


marcelsavegnago (approved):
LGTM

flotho (approved):
LGTM, make sense

pedrobaeza (commented):
Thanks for providing paths to get friendlier URLs. I think we can remove `-mgmt` prefix for shorter URLs. Do we also need a specific path for "My Tickets" instead of using general one?

ivantodorovich:
The thing is paths need to be unique.

I don't like it, but I added the `-mgmt` prefix to ensure we'll never conflict with Odoo's helpdesk paths. Although I guess it makes no sense to have both `helpdesk` and `helpdesk_mgmt` installed, so maybe it's pointless, but AFAICS it's not explicitly impossible.

For this reason, "My tickets" needs a specific path, too. Every action needs one, if we want a nice URL for it.

pedrobaeza:
We have already declared the incompatibility between this helpdesk and enterprise one, so let's remove the suffix, as we can't have a worse OCA solution because there's another one on enterprise.

ivantodorovich:
Alright then, it's done!

Thanks for taking the time to review, guys!

imlopes (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #814, opened 2025-08-13
[17.0][FIX] helpdesk_mgmt: Don't reset stage on user change

pedrobaeza:
Alternative solution to #811

Steps to reproduce:

- Create a ticket in a specific team.
- Change the stage to one that is not the first.
- Assign user of the same team.

Result: the stage is reset to the first one.

Analysis: Even if the team is not changed, the `_compute_stage_id` method is executed. We don't want to remove computed writable methods, as they are very useful, so let's just do a preventive check on the compute method for not changing the stage if the current one is still applicable for the current team.

It includes a test for avoiding future regressions.

@Tecnativa

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -19,8 +19,13 @@ class HelpdeskTicket(models.Model):
 
     @api.depends("team_id")
     def _compute_stage_id(self):
+        # This compute is executed on user change, even if not changing team, so let's
+        # apply a preventive check for not changing stage if the current one is still
+        # applicable to the current team
         for ticket in self:
-            ticket.stage_id = ticket.team_id._get_applicable_stages()[:1]
+            applicable_stages = ticket.team_id._get_applicable_stages()
+            if ticket.stage_id not in applicable_stages:
+                ticket.stage_id = applicable_stages[:1]
 
     @api.depends("team_id")
     def _compute_user_id(self):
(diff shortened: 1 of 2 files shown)


adasatorres:
@pedrobaeza, when different teams share the same stages, changing the team will not reset the ticket status.

pedrobaeza:
Yes, and honestly I think that's the correct thing (although not the subject of the specific correction). If you have already set that you are in a stage more advanced, why changing the team will reset the thing to the starting field?

adasatorres:
The workflow may be different, and not resetting it could create confusion for the newly assigned team.

pedrobaeza:
Let's see if others think this is conflicting. Do you have a lot of cases where ticket team is changed?

christian-ramos-tecnativa:
Im with @adasatorres on this one, changing the team should always trigger the stage reset (or maybe be configurable). What about setting up the stage reset in the write function? So we can compare the team_id with the previus one?

pedrobaeza:
@adasatorres @Christian-RB the "problem" is that dependencies are set statically when the registry is initialized, so for `user_id` change, the tree of affected fields is computed this way:

`user_id` > `team_id` > `stage_id`..

so the field `stage_id` is marked as modified, and recomputed even before `team_id` computed method is launched.

Why I don't want to go for a create/write override is because that changes are not reflected on live for the user, provoking confusion with things that are "magically" changed.

That's why I opted for this other solution, which is good enough IMO for most cases, with the only trade-off of a little behavior change in case you do team change (very infrequent).

Can we got with this?

christian-ramos-tecnativa (approved):
After some discussion with @adasatorres, we agree with your point that this approach will cover most cases. If a user wants to always reset the stage, a specific stages workflow can be created for each team as needed.

adasatorres (approved):

pedrobaeza:
Thanks for your understanding on this matter.

/ocabot merge patch

PR #817, opened 2025-08-14
[16.0][FIX] helpdesk_mgmt: Restrict ticket tag deletion

pedrobaeza:
If you have a ticket tag assigned in several tickets, and you remove the tag, it's lost on all the tickets you had it assigned, which leads to data lost.

Let's restrict the deletion of the tag if used through the standard mechanism in `ondelete` field attribute.

@Tecnativa TT57697


pilarvargas-tecnativa (approved):

carlos-lopez-tecnativa (commented):
LGTM

I think the `team_id` must have the restriction too.

flotho (approved):
Thanks,
make sense
Easy code review, 
LGTM

flotho:
> LGTM
> 
> I think the `team_id` must have the restriction too.

I think that this field is a many2one and the use case should be slightly different. Maybe another PR

PR #821, opened 2025-08-25
[18.0][FIX] helpdesk_mgmt: Helpdesk's portal tile must be always visible

### LINKED ISSUE #820: HelpDesk Menu in Account isn't showing
## Module
The name of the module that has a bug.

## Describe the bug
When accessing the customer portal (/my/tickets), tickets are listed correctly but there is no button to create a new ticket.
In many examples (including OCA’s own Runbot screenshots, e.g. issue #707), a "+ New" button is visible next to the "Tickets" menu, allowing portal users to create tickets directly.

## To Reproduce
**Affected versions**: 18.0

Install helpdesk_mgmt.
Log in as a portal user.
Navigate to /my/home.

Observe that no tickets button are available.

Image

**Expected behavior**
Portal users should be able to create new tickets directly from the portal, without requiring backend access or custom development.

ypapouin:
That's way, the portal's user can use the "New" button in the tile but also in the header if the user choose to browse existing tickets.

image

image

Fixing Issue #820

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_templates.xml
@@ -38,6 +38,9 @@
         inherit_id="portal.portal_docs_entry"
         primary="True"
     >
+        <xpath expr="//t[@t-set='force_show']" position="attributes">
+            <attribute name="t-value">True</attribute>
+        </xpath>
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Tested on runboat and working.

We can fast-track it for solving the issue.

/ocabot merge patch

Thanks for the patch, Yann.

PR #822, opened 2025-08-26
[16.0][IMP] helpdesk_mgmt: Improve usage of Kanban State

etobella:
Before this change, Kanban States was only showed on Kanban View and when we changed state, the value was not overriden (it happens in other places, like hr contract).

1. Now, it has been added a hook for setting it null when changing the state
2. Added in Form View 
   image
3. Added in Tree View 
   image
4. Added filters for kanban state


luisDIXMIT (approved):
Tested on runboat and LGTM!!

pedrobaeza (approved):

kluna1998 (approved):

PR #807, opened 2025-07-30
[17.0] [IMP] helpdesk_mgmt: categories visibility in portal

lperez-solvos:
This IMP introduce a category filter to control which one are displayed in helpdesk portal.
imagen
imagen

PR #824, opened 2025-08-27
[18.0] [IMP] helpdesk_mgmt: categories visibility in portal

lperez-solvos:
[fw-port it to 18] This IMP introduce a category filter to control which one are displayed in helpdesk portal.
17.0 PR  - https://github.com/OCA/helpdesk/pull/807


pedrobaeza (approved):

dalonsod:
Functional review, I've noticed this at filter view:

imagen


A separator between Archive and new filter should be added.

Could you check if v17 is also affected?

PR #829, opened 2025-09-02
[IMP] helpdesk_mgmt_project: Allow to use milestones

etobella:
Milestones are added by project but not used in the module. We just added them.

PR #831, opened 2025-09-03
[ADD] helpdesk_mgmt_timesheet_sale

etobella:
Depends on #829 

Allows to define sale order line directly on tickets, in a similar way that we have in PRoject tasks


helpdesk_mgmt_timesheet_sale/views/helpdesk_ticket.xml
```diff
@@ -0,0 +1,62 @@
...
+                    widget="so_line_field"
+                    groups="!sales_team.group_sale_salesman"
+                    column_invisible="not parent.allow_billable"
+                    readonly="readonly_timesheet"
+                    context="{'with_remaining_hours': True, 'with_price_unit': True}"
+                    options="{'no_create': True, 'no_open': True}"
+                    domain="[('is_service', '=', True), ('order_partner_id.commercial_partner_id.id', 'parent_of', parent.partner_id), ('is_expense', '=', False), ('state', '=', 'sale'), ('is_downpayment', '=', False)]"
+                    optional="hide"
+                />
+                <field
```
luisDIXMIT, line 39: It’s the same field with the same configuration, but one is visible to salespeople and the other to the rest of the groups. Is it correct?

luisDIXMIT (approved):
Code review and tested on runboat, LGTM. Only one minor comment.

helpdesk_mgmt_timesheet_sale/views/helpdesk_ticket.xml
etobella: probably not 😆

PR #833, opened 2025-09-09
[18.0] [FEAT] Init helpdesk_mgmt_notify_user

wouitmil:


helpdesk_mgmt_notify_user/data/mail_template.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <data noupdate="1">
+        <!--Email
+        template -->
+        <record id="mail_template_old_assign" model="mail.template">
+            <field name="name">Helpdesk : Notification à l'ancien assigné</field>
```
etobella: That should be in english

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -14,6 +14,7 @@ class HelpdeskTicket(models.Model):
         "mail.activity.mixin",
         "portal.mixin",
         "mail.tracking.duration.mixin",
+        "mail.thread.main.attachment",
```
etobella, line 17: Why do you need this here?

etobella:
You should use a single commit with the change and use the standard OCA nomenclature for commits please

PR #859, opened 2025-10-13
[18.0][IMP] helpdesk_mgmt: Improve user and team assignment in helpdesk tickets

carlos-lopez-tecnativa:
- Assign the current user by default if no user is set.
- Assign the first team of the user.
- When the team is changed, if the user is not in the team, unset the user.

Forward-port of https://github.com/OCA/helpdesk/pull/776


pedrobaeza (approved):
Fast-tracking for finishing the fw-port chain:

/ocabot merge patch

PR #870, opened 2025-10-29
[18.0][ADD] helpdesk_ticket_duplicate: New module

david-banon-tecnativa:
Adds a new module to allow marking tickets as duplicates.

@Tecnativa TT56371
@pedrobaeza @victoralmau Please review.

PR #872, opened 2025-10-29
[18.0][IMP] helpdesk_mgmt: Duplicate ticket tracking.

david-banon-tecnativa:
Moved https://github.com/OCA/helpdesk/pull/870 from a separate module into helpdesk_mgmt
Allows marking tickets as duplicates of another ticket moving them to a configurable stage.
This functionality is disabled by default.

@Tecnativa TT56371
@pedrobaeza @victoralmau @ivantodorovich  Please review.


ivantodorovich (approved):

pedrobaeza:
Thanks for the improvement. Tested on runboat, and I think several things can improve the UX:

- When marked as duplicated, put a ribbon like when you archived a record, putting "Duplicated".
- Instead of having a field in "Other information" tab, under the name, there should be a text "Duplicated of ..." and the many2one link to the ticket.

david-banon-tecnativa:
I think it looks better above the name and title. image

pedrobaeza:
OK, but let's put it as `info`, as this is not an "error" per se.

pedrobaeza (approved):
Great, thank you!

/ocabot merge minor

PR #874, opened 2025-10-30
[18.0][FIX] Ticket numbers received from mail

### LINKED ISSUE #853: helpdesk_mgmt sequence
When i create a new ticket on mail team incoming, sequence HT doesn't work.
Ticket number is the mail subject.
When i create manually a new ticket it works.

mileo:
FIX: [#mail](https://github.com/OCA/helpdesk/issues/853)

Correct ticket numbers that were incorrectly created from emails.

Changes include:

- The addition of an `x_number_backup` field to store the original ticket number value before correction.

- A migration script that:

- Backs up existing ticket numbers.

- Iterates over all tickets and corrects those whose number does not match the sequence format.

- The module version has been incremented to `18.0.1.13.2` to trigger the migration execution.


mathben:
I confirm, this fix the problem when receiving email, generate correctly the ticket number.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -55,6 +55,7 @@ def _read_group_stage_ids(self, stages, domain):
         return stages.search(search_domain)
 
     number = fields.Char(string="Ticket number", default="/", readonly=True)
+    x_number_backup = fields.Char(string="Number Backup")
```
flotho, line 58: Hi, 
Thanks for this proposal.
AFAIK, using prefix `x_` is not appropriate in the source code because Odoo will consider it as a manual field.
So this could be deleted from the UI.

flotho: Could you please consider using another field name ?
Thanks

flotho (changes requested):

christian-ramos-tecnativa:
I’m not fully confident about the `pre-migration.py` script because it assumes the original prefix hasn’t changed. I can imagine a scenario where a user changes the prefix but leaves the previous records untouched.
In my opinion, each case should be handled individually. So you woudln't need the new field.
Adding the default to the `message_new` method works for me thanks

pedrobaeza:
Superseded by #895

PR #876, opened 2025-10-30
[16.0][IMP] helpdesk_mgmt: Extends PortalHomeCounters to include "ticket_count" in the always displayed counters

marcelsavegnago:
Extends PortalHomeCounters to include "ticket_count" in the always displayed counters

cc @parzewski


kaynnan (approved):

pedrobaeza (approved):
Tested and working. Thanks for the patch.

PR #877, opened 2025-10-31
[16.0][REF] helpdesk_portal_restriction: call super().create_new_ticket() and override only categories in qcontex

marcelsavegnago:
This pull request refactors the `create_new_ticket` method in the `helpdesk_portal_restriction/controllers/main.py` file to simplify its logic and delegate more responsibility to the parent class. The main change is to call the superclass implementation and then inject the custom categories into the response context, instead of manually constructing the entire context.

Refactoring and code simplification:

* Refactored the `create_new_ticket` method to call `super().create_new_ticket(**kw)` and then update the response context with custom categories, reducing code duplication and leveraging the parent class's logic.

Cleanup:

* Removed unused import of `request` from `odoo.http` to tidy up the imports.

### DIFF
--- a/helpdesk_portal_restriction/controllers/main.py
+++ b/helpdesk_portal_restriction/controllers/main.py
@@ -1,7 +1,6 @@
 import logging
 
 import odoo.http as http
-from odoo.http import request
 
 from odoo.addons.helpdesk_mgmt.controllers.main import HelpdeskTicketController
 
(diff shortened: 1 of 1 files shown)


kaynnan (approved):

lbarry-apsl (approved):
Code review and runboat, it's correct!

PR #882, opened 2025-11-10
[18.0][IMP] helpdesk_mgmt_sla: Add company_id to helpdesk.sla

david-banon-tecnativa:
Adds a company_id field to SLAs, with False as a default to keep previous workflow intact.

@Tecnativa TT56371


david-banon-tecnativa:
Review please @pedrobaeza @eduezerouali-tecnativa

helpdesk_mgmt_sla/models/helpdesk_sla.py
```diff
@@ -11,6 +11,11 @@ class HelpdeskSla(models.Model):
     _description = "Helpdesk SLA"
 
     name = fields.Char(required=True)
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        string="Company",
+        default=False,
```
pedrobaeza: This is the default, so it's redundant.

helpdesk_mgmt_sla/views/helpdesk_sla_views.xml
```diff
@@ -34,6 +34,12 @@
                         </h1>
                     </div>
                     <group>
+                        <field
+                            name="company_id"
+                            placeholder="All Companies"
```
pedrobaeza:
```suggestion
                            placeholder="Empty for all the companies"
```

helpdesk_mgmt_sla/views/helpdesk_sla_views.xml
```diff
@@ -81,6 +87,7 @@
                 <field name="days" />
                 <field name="hours" />
                 <field name="note" />
+                <field name="company_id" optional="hide" />
```
pedrobaeza: Put the groups here and I wouldn't put it `optional`.

pedrobaeza (approved):

christian-ramos-tecnativa (approved):
Non blocking comment, but IMO it would be nice to add company restriction between SLA and Teams when the company is selected

pedrobaeza:
Yeah, good point. Please do it, @david-banon-tecnativa. Check https://www.odoo.com/documentation/18.0/es/developer/howtos/company.html

david-banon-tecnativa:
But adding the company check as Odoo recommends will cause previous correct workflows to be invalid once merged, ex:
Previously, i have SLA 1 assigned to teams of company A and B, now SLA 1 has company_id = False, and it can only be applied on teams also with `company_id = False`, i think this should be left for a v19 migration, and here a domain to restrict company A to company B relations should be added.

christian-ramos-tecnativa:
The logic should be:
- If SLA `company_id = False` then any team can be assign
- If SLA has `company_id` just teams with the same one or none can be assign

I think you could add a domain to the field like:
`['|', (0 if company_id else 1, "=", 1), '|', '('company_id', '=', False), ('company_id', '=', company_id)]`

PR #883, opened 2025-11-11
[18.0][IMP] helpdesk_mgmt: Update child team and category translations

david-banon-tecnativa:
All other helpdesk.ticket.* names area already translatable, so no reason to leave helpdesk.ticket.team behind.

@christian-ramos-tecnativa @pedrobaeza 
@Tecnativa TT56371

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket_category.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket_category.py
@@ -34,11 +34,20 @@ class HelpdeskCategory(models.Model):
     )
     parent_path = fields.Char(index=True)
     complete_name = fields.Char(
-        compute="_compute_complete_name", store=True, recursive=True
+        compute="_compute_complete_name",
+        recursive=True,
+        search="_search_complete_name",
     )
     show_in_portal = fields.Boolean(default=True)
 
+    def _search_complete_name(self, operator, value):
+        records = self.search_fetch([], ["complete_name"]).filtered_domain(
+            [("complete_name", operator, value)]
+        )
+        return [("id", "in", records.ids)]
+
     @api.depends("name", "parent_id.complete_name")
+    @api.depends_context("lang")
     def _compute_complete_name(self):
         for category in self:
             if category.parent_id:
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
This can't be then stored:

https://github.com/Tecnativa/helpdesk/blob/df3697ceb9f31456f26422b01123a6fb55986d74/helpdesk_mgmt/models/helpdesk_ticket_team.py#L74

as it's going to store only one language. Check if we can pass it to non-storable.

david-banon-tecnativa:
Updated compute methods to also compute complete_name in all installed langs.

pedrobaeza (approved):
Seems to be working this way in multi-language mode!

christian-ramos-tecnativa:
I think there may be something wrong on how the recursive attribute works, this is my test (in runboat):
- Team structure:
image

- Changing the father name doesn't recompute the child complete name:
image

- Changing the grandfather name updates the `complete_name` in father and now the child `complete_name` contains the father name changed but not the grandfather update:
image

Extra information:
The `Localization team` was already created, but the father and grandfather are new. Is it posible that the compute is procesing the records in the wrong order?

marcelsavegnago (approved):
LGTM

helpdesk_mgmt/models/helpdesk_ticket_category.py
```diff
@@ -34,11 +34,21 @@ class HelpdeskCategory(models.Model):
     )
     parent_path = fields.Char(index=True)
     complete_name = fields.Char(
-        compute="_compute_complete_name", store=True, recursive=True
+        compute="_compute_complete_name",
+        recursive=True,
+        translate=True,
```
pedrobaeza: I think you don't need this anymore.

helpdesk_mgmt/models/helpdesk_ticket_category.py
```diff
@@ -34,11 +34,21 @@ class HelpdeskCategory(models.Model):
     )
     parent_path = fields.Char(index=True)
     complete_name = fields.Char(
-        compute="_compute_complete_name", store=True, recursive=True
+        compute="_compute_complete_name",
+        recursive=True,
```
pedrobaeza: And the same for this.

helpdesk_mgmt/models/helpdesk_ticket_category.py
pedrobaeza: ping @david-banon-tecnativa

pedrobaeza (approved):
Changed the approach to be non stored with a search method, for avoiding the issue notified by Christian

/ocabot merge minor

PR #886, opened 2025-11-18
[16.0][IMP] helpdesk_mgmt: separate filters to force AND logic

NL66278:
When filters are not separated with another element, their domains are combined with an OR selection. This is almost certainly not what you want. If you select Open Tickets end My tickets, you expect your tickets that are also open, not a combination of open tickets and tickets assigned to you.

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_views.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_views.xml
@@ -33,12 +33,15 @@
                     name="unassigned"
                     domain="[('user_id','=',False)]"
                 />
+                <separator />
                 <filter string="Open" name="open" domain="[('closed', '=', False )]" />
+                <separator />
                 <filter
                     string="Unattended"
                     name="unattended"
                     domain="[('unattended', '=', True )]"
                 />
+                <separator />
                 <filter
                     string="High Priority"
                     name="high_priority"
@@ -73,6 +76,7 @@
                     name="activities_my"
                     domain="[('activity_ids.user_id', '=', uid)]"
                 />
+                <separator />
                 <filter
                     string="Late Activities"
                     name="activities_overdue"


dreispt (approved):

CristianoMafraJunior (approved):

marcelsavegnago (approved):

helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -73,34 +77,26 @@
...
                     domain="[('activity_ids.user_id', '=', uid)]"
                 />
+                <separator />
                 <filter
                     string="Late Activities"
                     name="activities_overdue"
                     domain="[('activity_ids.date_deadline', '&lt;', context_today().strftime('%Y-%m-%d'))]"
                     help="Show all records which has next action date is before today"
                 />
+                <separator />
```
pedrobaeza: This separator shouldn't be.

helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -56,6 +59,7 @@
                     name="mytickets"
                     domain="[('user_id','=',uid)]"
                 />
+                <separator />
```
pedrobaeza: I think you may also want to join your tickets and your followed tickets at the same time, so please remove this separator as well.

helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -73,6 +77,7 @@
                     name="activities_my"
                     domain="[('activity_ids.user_id', '=', uid)]"
                 />
+                <separator />
```
pedrobaeza, line 79: And you also want to join by activities, so I don't think this one should be.

pedrobaeza (commented):
Sorry, reviewing deeply, I don't think we need so many separators.

helpdesk_mgmt/views/helpdesk_ticket_views.xml
NL66278: Removed this one as well

NL66278:
@pedrobaeza I think with the set of separators presented now there is a logical subdivision between filter criteria you might want to OR (like my tickets and my followed tickets), and criteria you likely want to AND (My tickets and open Tickets, resulting in My open tickets).

pedrobaeza (approved):
/ocabot merge patch

PR #889, opened 2025-11-27
[17.0][FIX] helpdesk_mgmt: Do not automatically assign a user when creating tickets from the portal

victoralmau:
Do not automatically assign a user when creating tickets from the portal

Previously, the portal user who created the ticket was defined as the ticket user (completely incorrect)

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT59493

PR #890, opened 2025-11-27
[17.0][FIX] helpdesk_mgmt: Do not self-define user in tickets

victoralmau:
Do not self-define user in tickets

Previously, when creating a ticket, the user who was creating it was defined as the user, even if no team was defined; now, that logic has been removed from `_compute_user_id()` and moved a configuration option that allows each
company to decide whether tickets are self-assigned or not.

It is important to clarify that this change is **NOT** added to https://github.com/OCA/helpdesk/pull/889 because, although they are similar, they are different, and this change will probably generate debate and take time to merge.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT59493

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -30,8 +30,6 @@ class HelpdeskTicket(models.Model):
     @api.depends("team_id")
     def _compute_user_id(self):
         for ticket in self:
-            if not ticket.user_id:
-                ticket.user_id = self.env.user
             if ticket.team_id and ticket.user_id not in ticket.team_id.user_ids:
                 # If the user is not part of the team, we remove the user
                 ticket.user_id = False
@@ -155,6 +153,26 @@ class HelpdeskTicket(models.Model):
         help="Gives the sequence order when displaying a list of tickets.",
     )
     active = fields.Boolean(default=True)
+
+    @api.model
+    def default_get(self, fields):
+        # The appropriate user is defined only if the "Auto assign User" option is
+        # checked in the company.
+        # If the team is set, the user must belong to that team.
+        defaults = super().default_get(fields)
+        company_id = defaults.get("company_id") or self.env.company.id
+        if "user_id" in fields and not defaults.get("user_id"):
+            company = self.env["res.company"].browse(company_id)
+            if company.helpdesk_mgmt_ticket_auto_assign:
+                if defaults.get("team_id"):
+                    team = self.env["helpdesk.ticket.team"].browse(
+                        defaults.get("team_id")
+                    )
+                    if self.env.user in team.user_ids:
+                        defaults["user_id"] = self.env.user.id
+                else:
+                    defaults["user_id"] = self.env.user.id
+        return defaults
 
     @api.depends("name")
     def _compute_display_name(self):
--- a/helpdesk_mgmt/views/res_config_settings_views.xml
+++ b/helpdesk_mgmt/views/res_config_settings_views.xml
@@ -60,6 +60,25 @@
                             </div>
                         </setting>
                     </block>
+                    <block title="Helpdesk">
+                        <setting
+                            string="Tickets"
+                            id="helpdesk_mgmt_ticket_misc"
+                            help="Controls if the tickets created are automatically assigned to the user."
+                        >
+                            <div class="mt16">
+                                <div class="content-group">
+                                    <div>
+                                        <field
+                                            name="helpdesk_mgmt_ticket_auto_assign"
+                                            string="Auto assign User"
+                                        />
+                                        <label for="helpdesk_mgmt_ticket_auto_assign" />
+                                    </div>
+                                </div>
+                            </div>
+                        </setting>
+                    </block>
                 </app>
             </xpath>
         </field>
--- a/helpdesk_mgmt/models/res_company.py
+++ b/helpdesk_mgmt/models/res_company.py
@@ -18,3 +18,7 @@ class Company(models.Model):
         string="Required Category field in Helpdesk portal",
         default=True,
     )
+    helpdesk_mgmt_ticket_auto_assign = fields.Boolean(
+        string="Auto assign tickets",
+        default=True,
+    )
--- a/helpdesk_mgmt/models/res_config_settings.py
+++ b/helpdesk_mgmt/models/res_config_settings.py
@@ -18,3 +18,7 @@ class ResConfigSettings(models.TransientModel):
         related="company_id.helpdesk_mgmt_portal_category_id_required",
         readonly=False,
     )
+    helpdesk_mgmt_ticket_auto_assign = fields.Boolean(
+        related="company_id.helpdesk_mgmt_ticket_auto_assign",
+        readonly=False,
+    )
--- a/helpdesk_mgmt/tests/test_helpdesk_ticket.py
+++ b/helpdesk_mgmt/tests/test_helpdesk_ticket.py
@@ -213,6 +213,32 @@ class TestHelpdeskTicket(TestHelpdeskTicketBase):
             new_ticket_form.team_id = new_team
             self.assertFalse(new_ticket_form.user_id)
 
+    def test_ticket_auto_assign(self):
+        self.company.helpdesk_mgmt_ticket_auto_assign = True
+        ticket_a = (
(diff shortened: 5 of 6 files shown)


pedrobaeza:
I think we can set the user and no team. When the team is set, then the coherence is checked.

victoralmau:
> I think we can set the user and no team. When the team is set, then the coherence is checked.

That is not correct and would also have side effects; the team defines itself as https://github.com/OCA/helpdesk/blob/41eb3e6e9a4467f7e53b32e1a195c50366a8f706/helpdesk_mgmt/models/helpdesk_ticket.py#L169

I have come up with another approach that IMO works for all cases:
- Users who have been using the helpdesk until now will continue to have the same behavior, and tickets will still be auto-assigned when a new one is created.
- Users who want to decide whether or not tickets are auto-assigned when created can do so in the settings.

What do you think, @carlos-lopez-tecnativa ?

helpdesk_mgmt/controllers/main.py
```diff
@@ -105,7 +105,8 @@ def _prepare_submit_ticket_vals(self, **kw):
     @http.route("/submitted/ticket", type="http", auth="user", website=True, csrf=True)
     def submit_ticket(self, **kw):
         vals = self._prepare_submit_ticket_vals(**kw)
-        new_ticket = request.env["helpdesk.ticket"].sudo().create(vals)
+        ticket_model = request.env["helpdesk.ticket"].sudo()
+        new_ticket = ticket_model.with_context(skip_auto_assign=True).create(vals)
```
pedrobaeza: Wasn't this case already covered passing `user_id=False` in the vals dictionary?

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -205,6 +203,13 @@ def create(self, vals_list):
                 )
                 if channel_email_id:
                     vals["channel_id"] = channel_email_id.id
+            # Auto-assign user
+            if "user_id" not in vals:
+                defaults = self.default_get(["company_id"])
+                company_id = vals.get("company_id", defaults.get("company_id"))
+                company = self.env["res.company"].browse(company_id)
+                if company.helpdesk_mgmt_ticket_auto_assign:
+                    vals["user_id"] = self.env.user.id
```
carlos-lopez-tecnativa: I placed this code before the other code so that `assigned_date` works.
Or do you not consider that assigning the user automatically should set the `assigned_date`?
I think it should.

image

carlos-lopez-tecnativa (changes requested):
This is not working as expected. I create a new ticket and save, but the user is not assigned. Note that I did not select a team.
image

And the feature is enabled in the settings:
image

Before this commit, the user was displayed as assigned because the compute was triggered by default. Now I must save the record to see the assigned user. This can cause confusion for the user: first they see the field empty and later they see the assigned user. I think this should be done in the compute method, considering the company setting, instead of doing it in the create method.

victoralmau:
> This is not working as expected. I create a new ticket and save, but the user is not assigned. Note that I did not select a team. image
> 
> And the feature is enabled in the settings: image
> 
> Before this commit, the user was displayed as assigned because the compute was triggered by default. Now I must save the record to see the assigned user. This can cause confusion for the user: first they see the field empty and later they see the assigned user. I think this should be done in the compute method, considering the company setting, instead of doing it in the create method.

I believe that the new approach takes all cases into account.

helpdesk_mgmt/views/res_config_settings_views.xml
```diff
@@ -60,6 +60,25 @@
...
+                            string="Tickets"
+                            id="helpdesk_mgmt_ticket_misc"
+                            help="Controls if the tickets created are automatically assigned to the user."
+                        >
+                            <div class="mt16">
+                                <div class="content-group">
+                                    <div>
+                                        <field
+                                            name="helpdesk_mgmt_ticket_auto_assign"
+                                            string="Auto assign"
```
carlos-lopez-tecnativa:
```suggestion
                                            string="Auto assign User"
```

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -30,8 +30,6 @@ def _compute_stage_id(self):
     @api.depends("team_id")
     def _compute_user_id(self):
         for ticket in self:
-            if not ticket.user_id:
-                ticket.user_id = self.env.user
             if ticket.team_id and ticket.user_id not in ticket.team_id.user_ids:
                 # If the user is not part of the team, we remove the user
                 ticket.user_id = False
```
carlos-lopez-tecnativa, line 35: Your changes are working now, but in this PR https://github.com/OCA/helpdesk/pull/776 I cover the following case:

`When the team is changed, if the user is not in the team, unset the user.`

This works halfway:

- When I change the team and the user is not a member, the user is unset — working fine.
- But when I go to tickets from the dashboard and the team is set by default_team_id, if the user is not a member of the team, the user must not be assigned. With this PR the user is assigned, and this is not expected. Please consider this case. If you need more information, please let me know.

`Marc Demo` is a member of the `Localization team`.
`Mitchell Admin` is a member of `Consultants`.

So when I enter from the dashboard through the `Localization team`, the user `Mitchell Admin` must not be set as the user, because this user is not a member of that team.

Before this PR


https://github.com/user-attachments/assets/e827ffdf-e23d-47a8-a971-5949cfced56c


After this PR


https://github.com/user-attachments/assets/808f0f3b-5fba-4c35-8e28-f2a666c30542

carlos-lopez-tecnativa (changes requested):

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -182,6 +197,11 @@ def create(self, vals_list):
         for vals in vals_list:
             if vals.get("number", "/") == "/":
                 vals["number"] = self._prepare_ticket_number(vals)
+            # Auto-assign user
+            if "user_id" not in vals:
+                defaults = self.default_get(["user_id"])
+                if defaults.get("user_id"):
+                    vals["user_id"] = self.env.user.id
```
carlos-lopez-tecnativa: Why not? It’s the same but more intuitive.

```suggestion
                if defaults.get("user_id"):
                    vals["user_id"] = defaults.get("user_id")
```

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -156,6 +154,23 @@ def _read_group_stage_ids(self, stages, domain, order):
     )
     active = fields.Boolean(default=True)
 
+    @api.model
+    def default_get(self, fields):
+        defaults = super().default_get(fields)
+        company_id = defaults.get("company_id") or self.env.company.id
+        if not defaults.get("user_id"):
```
carlos-lopez-tecnativa: To avoid doing it if it’s not necessary.
```suggestion
        if "user_id" in fields and not defaults.get("user_id"):
```

carlos-lopez-tecnativa (changes requested):
LGTM, just a few minor changes to improve the code.

carlos-lopez-tecnativa (approved):
LGTM

rousseldenis (approved):
Great!

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -156,6 +154,23 @@ def _read_group_stage_ids(self, stages, domain, order):
     )
     active = fields.Boolean(default=True)
 
+    @api.model
+    def default_get(self, fields):
+        defaults = super().default_get(fields)
```
pedrobaeza, line 162: Add comment summarizing what this override does.

helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -182,6 +197,11 @@ def create(self, vals_list):
         for vals in vals_list:
             if vals.get("number", "/") == "/":
                 vals["number"] = self._prepare_ticket_number(vals)
+            # Auto-assign user
+            if "user_id" not in vals:
+                defaults = self.default_get(["user_id"])
```
pedrobaeza: Is this a real case? Shouldn't the default always be executed?

pedrobaeza (approved):
/ocabot merge minor

PR #894, opened 2025-12-02
[16.0][IMP] helpdesk_mgmt: make partner fields computed

nicolas-delbovier-acsone:
- Use a compute method for 'partner_name' and 'partner_email' based on partner_id
- Make the fields stored and not readonly so that users can still enter an email or a name if the partner is still not registered in odoo


sbejaoui (approved):

hoangtrann:
I have gone this path and would say that it's best to get the value automatically from partner and fill to the helpdesk ticket. If user make changes, display a hint that they can update the contact as well if that's what they want.

There are a lot of scenario that customer support want to just change the name or email on the ticket and not having them affect the actual contact.

lmignon (approved):

marcelsavegnago (approved):
LGTM

WesleyOliveira98 (approved):
CODE REVIEW - OK

PR #896, opened 2025-12-03
[18.0][FIX] helpdesk_mgmt_fieldservice: Test, remove is_button

christian-ramos-tecnativa:
`is_button` doesn't exists since https://github.com/OCA/field-service/pull/1471 and the way html is compared changed upstream

@Tecnativa TT59622
@pedrobaeza @victoralmau


pedrobaeza (approved):
Let's fix the repo

/ocabot merge patch

PR #897, opened 2025-12-04
[18.0][FIX] helpdesk_mgmt: Do not self-define user in tickets

victoralmau:
FWP from 17.0: https://github.com/OCA/helpdesk/pull/890

Do not self-define user in tickets

Previously, when creating a ticket, the user who was creating it was defined as the user, even if no team was defined; now, that logic has been removed from `_compute_user_id()` and moved a configuration option that allows each
company to decide whether tickets are self-assigned or not.

It is important to clarify that this change is **NOT** added to https://github.com/OCA/helpdesk/pull/889 because, although they are similar, they are different, and this change will probably generate debate and take time to merge.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT59493

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -25,8 +25,6 @@ class HelpdeskTicket(models.Model):
     @api.depends("team_id")
     def _compute_user_id(self):
         for ticket in self:
-            if not ticket.user_id:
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):
Code review

carlos-lopez-tecnativa (approved):

PR #909, opened 2025-12-29
[16.0][IMP] helpdesk_motive: allow selecting reasons not assigned to a team

sbejaoui:
in practice, some reasons are generic and not specific to any helpdesk team

this PR extends the field domain to also allow selecting reasons that are not assigned to any team


lmignon (approved):

marcelsavegnago (approved):
LGTM

PR #914, opened 2026-01-07
[18.0][FIX] helpdesk_mgmt: add teams to portal domain

christian-ramos-tecnativa:
Before this PR whenever you create a ticket with a portal user as contact and assign a team without `show_in_portal`, the next error raise, not allowing you to see any ticket in `/my/tickets`.
image
The solution proposed here is filtering out non visible team tickets.
@Tecnativa TT60091
@pedrobaeza @david-banon-tecnativa @victoralmau

### DIFF
--- a/helpdesk_mgmt/controllers/myaccount.py
+++ b/helpdesk_mgmt/controllers/myaccount.py
@@ -104,6 +104,13 @@ class CustomerPortalHelpdesk(CustomerPortal):
             ]
         )
 
+        domain = AND(
+            [
+                domain,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But the team visibility is just for ticket creation, not for viewing tickets with that team assigned, so I think the record rules shouldn't restrict in that sense.

christian-ramos-tecnativa:
I implemented it this way to preserve the ability to create tickets for users you don’t want to display in the portal

pedrobaeza (approved):
This sounds better.

victoralmau (approved):

PR #917, opened 2026-01-09
[18.0] Revert "[FIX] helpdesk_mgmt: add teams to portal domain"

christian-ramos-tecnativa:
This reverts https://github.com/OCA/helpdesk/pull/914 as the error was wrong diagnosed
@Tecnativa TT60091
@pedrobaeza @victoralmau


pedrobaeza (approved):
OK, let's fast-track it for reverting to the original situation. I didn't see it coming also.

/ocabot merge patch

PR #918, opened 2026-01-09
[18.0][FIX] helpdesk_mgmt_timesheet: access error to team_id

christian-ramos-tecnativa:
I misinterpreted the error in https://github.com/OCA/helpdesk/pull/914, as it was not related to the base module but to this one. The fix has been applied, and a reversion is proposed in https://github.com/OCA/helpdesk/pull/917

@Tecnativa TT60091
@pedrobaeza @victoralmau


pedrobaeza (approved):
Now it makes more sense:

/ocabot merge patch

PR #592, opened 2024-06-17
[14.0] [FIX] helpdesk_mgmt_fieldservice: improve _compute_all_closed method

renda-dev:
proposing again https://github.com/OCA/helpdesk/pull/454

(changes made to `helpdesk_mgmt_fieldservice/static/description/index.html` are from pre-commit)

@OCA/helpdesk-maintainers 

---

Checking a boolean field is more reliable than checking arbitrary names that can be changed by the user

PR #920, opened 2026-01-13
[16.0][IMP] helpdesk_mgmt_fieldservice: simplify all_orders_closed

SirPyTech:
Backporting from `18.0` (https://github.com/OCA/helpdesk/commit/6cbe7ebfb3c47983d4515523e0713a12ec192643) the fine work that @ivantodorovich has done on the field `helpdesk.ticket.all_orders_closed` :clap:

This also includes (actually, overwrites) the fix https://github.com/OCA/helpdesk/pull/592 that was merged in `14.0` but not ported to `16.0` in the migration https://github.com/OCA/helpdesk/pull/608.

@ivantodorovich I slightly edited your commit to adapt it for `16.0`, please let me know if that's ok.


helpdesk_mgmt_fieldservice/views/helpdesk_ticket_views.xml
```diff
@@ -71,7 +70,7 @@
             <filter name="open" position="after">
                 <filter
                     string="Closed FSM Orders"
-                    domain="[('stage_id.closed', '=', False), ('all_orders_closed', '=', True)]"
+                    domain="[('stage_id.closed', '=', False), ('fsm_order_ids', 'any', [('stage_id.is_closed', '=', False)])]"
```
ivantodorovich:
```suggestion
                    domain="[('stage_id.closed', '=', False), ('fsm_order_ids', '!=', False), ('fsm_order_ids.stage_id.is_closed', '=', False)]"
```

In 16.0 we don't have the `any` operator yet, but I think this should be the same

ivantodorovich (approved):
Thanks @SirPyTech !

One small comment about the `any` operator, not yet available in 16.0.
Otherwise LGTM :) 

Pre-approving

helpdesk_mgmt_fieldservice/views/helpdesk_ticket_views.xml
SirPyTech: Thanks!
I wasn't sure this was supported in `16.0` but was reassured because it is used in https://github.com/odoo/odoo/blob/70dd6b6cc5e43370295c98ee7dc9ab11b7f1a387/addons/l10n_it_edi_doi/views/res_partner_views.xml#L12.

I double checked and you are correct: it is not there in https://github.com/odoo/odoo/blob/70dd6b6cc5e43370295c98ee7dc9ab11b7f1a387/addons/web/static/src/legacy/js/core/domain.js.

So I applied your suggestion, please check :slightly_smiling_face:

ivantodorovich (approved):

helpdesk_mgmt_fieldservice/views/helpdesk_ticket_views.xml
ivantodorovich: Awesome, thanks!

marcelsavegnago (approved):
LGTM

BhaveshHeliconia (approved):
LGTM!

SirPyTech:
> /ocabot merge minor

Thanks! :blush:

PR #933, opened 2026-01-26
[18.0][FIX] helpdesk_mgmt_timesheet: Fix With activity today filter

christian-ramos-tecnativa:
Right now the filter "With activity today" is raising an error when trying to access` datetime.datetime`
@Tecnativa TT60514
@victoralmau @david-banon-tecnativa

### DIFF
--- a/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
+++ b/helpdesk_mgmt_timesheet/views/helpdesk_ticket_view.xml
@@ -24,7 +24,7 @@
                     name="with_activity_today"
                     string="With activity today"
                     groups="hr_timesheet.group_hr_timesheet_user"
-                    domain="[('last_timesheet_activity', '=', datetime.datetime.today())]"
+                    domain="[('last_timesheet_activity', '=', context_today())]"
                 />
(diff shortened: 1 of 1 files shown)


victoralmau:
The appropriate change is probably the same as in v17: https://github.com/OCA/helpdesk/pull/783

christian-ramos-tecnativa:
@victoralmau Do you know by any chance why @pedrobaeza did it that way? It looks like it was overcomplicated, because the domain will be `(last_timesheet_activity > today-1) AND (last_timesheet_activity <= today)` isn't that strictly equal to  `(last_timesheet_activity == today)`? At least on this version works as intended

pedrobaeza:
Afaik it's a datetime, so it wasn't working today expression. Maybe that had been improved...

christian-ramos-tecnativa:
The field `last_timesheet_activity` is a Date so this approach should be better. Just to be sure, I tried with a timesheet datetime in the future but same day and it was properly displayed.

pedrobaeza (approved):
Great then!

victoralmau (approved):
Code review OK

PR #940, opened 2026-02-05
[17.0][FIX] helpdesk_ticket_partner_response: Change stage on fetchmail

pedrobaeza:
When receiving mails from fetchmail, the comparison done about the partner equality makes that the stage change never happens, as the user is not the customer, but the OdooBot or cron user.

Let's remove then such comparison.

@Tecnativa TT60576


peluko00 (approved):
LGTM!

/ocabot merge patch

PR #946, opened 2026-02-11
[18.0][FIX] helpdesk_mgmt_project: project not associate at task creation

mathben:
The project is not associate when create ticket without modification. The context not working.
A solution is to disable quick creation. Or maybe we can fix it without disable it?

The option `Create "Test creation"` not working, only `Create and edit...`
This commit disable first option
image

This is the error when selection `Create "Test creation"`
Project is missing
image

### DIFF
--- a/helpdesk_mgmt_project/views/helpdesk_ticket_view.xml
+++ b/helpdesk_mgmt_project/views/helpdesk_ticket_view.xml
@@ -41,6 +41,7 @@
                     domain="[('project_id', '=', project_id)]"
                     invisible="not project_id"
                     context="{'default_project_id': project_id}"
+                    options="{'no_quick_create': True}"
                     groups="project.group_project_user"
                 />
                 <t groups="project.group_project_milestone">


flotho (commented):
Hi,

Does it happened during the creation process of the ticket or once it has been saved?

mathben:
@flotho 
> Does it happened during the creation process of the ticket or once it has been saved?

Once it has been saved, I didn't test at creation process.

I believe this pull request works around a hidden bug by disabling the functionality.

flotho:
> @flotho
> 
> > Does it happened during the creation process of the ticket or once it has been saved?
> 
> Once it has been saved, I didn't test at creation process.
> 
> I believe this pull request works around a hidden bug by disabling the functionality.

Thanks @mathben , I think that the community will better expect that you foind the hidden bug ;-)
Any chance to test same behavior on a fresh OCB install with minimal modules and other objects ?

Regards

PR #953, opened 2026-03-02
[17.0][FIX] helpdesk_mgmt_rating: Fix translations

victoralmau:
Fix translations

It seems that defining `<b t-out=“object.stage_id.name” />` in the template causes problems with translations because it repeats the code several times.

plantilla-antes

This does not happen in v18. To fix it, the same approach has been used in `project` (https://github.com/odoo/odoo/blob/24b2b5879ea700ad88d33ee0a85a84deb03fbdba/addons/project/data/mail_template_data.xml#L90) and the translation has been modified directly to confirm that this code is not repeated.

plantilla-despues

@Tecnativa TT61157


christian-ramos-tecnativa (approved):

pedrobaeza (changes requested):
es.po contains too many changes that shouldn't be. This may be because it has been edited with POEdit.

luisDIXMIT (approved):
Tested on runboat and LGTM!

pedrobaeza (approved):
/ocabot merge patch

PR #955, opened 2026-03-05
[17.0][IMP] helpdesk_mgmt: allows team leader to follow tickets automatically

cgarcia-solvos:
Add an optional setting in helpdesk teams to automatically subscribe the team leader as a follower of tickets belonging to that team.

imagen

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket_team.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket_team.py
@@ -75,6 +75,11 @@ class HelpdeskTeam(models.Model):
         compute="_compute_complete_name", store=True, recursive=True
     )
     parent_path = fields.Char(index=True, unaccent=False)
+    add_leader_as_follower = fields.Boolean(
+        string="Leader follows tickets",
+        default=False,
+        help="If enabled, the team leader will automatically follow new tickets.",
+    )
 
     @api.depends("name", "parent_id.complete_name")
     def _compute_complete_name(self):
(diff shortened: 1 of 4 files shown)


helpdesk_mgmt/views/helpdesk_ticket_team_views.xml
```diff
@@ -102,6 +102,7 @@
                                 groups="base.group_multi_company"
                             />
                             <field name="show_in_portal" />
+                            <field name="add_leader_as_follower" />
```
dalonsod, line 105: Could you move this option immediately below team leader field, and make it visible only if team leader is set?

helpdesk_mgmt/models/helpdesk_ticket_team.py
```diff
@@ -75,6 +75,11 @@ class HelpdeskTeam(models.Model):
         compute="_compute_complete_name", store=True, recursive=True
     )
     parent_path = fields.Char(index=True, unaccent=False)
+    add_leader_as_follower = fields.Boolean(
```
dalonsod, line 78: This field could be computed, in order to unset it when team leader field is unset as well

dalonsod (changes requested):
See minor comments

pedrobaeza:
Why not using the standard mechanism through followers in the team?

cgarcia-solvos:
@pedrobaeza Thanks for pointing that out.

You are right, we hadn't noticed that behavior.
Even so, in our use case this still requires a manual step, since the team leader needs to be explicitly added as a follower of each team.

What we were actually trying to achieve is to automate this behavior. One possible alternative could be introducing a company-level setting such as:

“Team leaders automatically follow their teams”

With this option enabled, the team leader would automatically be added as a follower of the team.

What do you think about this approach?

pedrobaeza:
Then this should be put as an extra module, not inside the main one, for not complicating the base with a lot of options, checks, code, etc.

pedrobaeza:
I also wonder how many teams/leaders you have for being a barrier to do this setup once, and instead, wanting to automate it this way.

PR #964, opened 2026-03-11
[18.0][FIX] helpdesk_ticket_partner_response: Error on New Ticket

### LINKED ISSUE #963: Error on New Ticket with helpdesk_ticket_partner_response
1) When we receive an New Support Ticket via Email 

We get this Error:

Traceback (most recent call last):
  File "/opt/odoo/auto/addons/mail/models/fetchmail.py", line 238, in fetch_mail
    res_id = MailThread.with_context(**additionnal_context).message_process(server.object_id.model, data[0][1], save_original=server.original, strip_attachments=(not server.attach))
  File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 1296, in message_process
    thread_id = self._message_route_process(message, msg_dict, routes)
  File "/opt/odoo/auto/addons/mass_mailing/models/mail_thread.py", line 28, in _message_route_process
    return super(MailThread, self)._message_route_process(message, message_dict, routes)
  File "/opt/odoo/auto/addons/helpdesk_ticket_partner_response/models/mail_thread.py", line 9, in _message_route_process
    self.change_status_ticket_from_portal(routes)
  File "/opt/odoo/auto/addons/helpdesk_ticket_partner_response/models/mail_thread.py", line 15, in change_status_ticket_from_portal
    ticket = self.env["helpdesk.ticket"].sudo().browse(int(ticket_id))
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

https://github.com/OCA/helpdesk/blob/cbeab76c97b12acfcfeced5fbedef15be0b84cdd/helpdesk_ticket_partner_response/models/mail_thread.py#L15

2) It should be possible to Reset Tickets that are not Assgined to a Particular Team

CristianoMafraJunior:
Fixes #963

## Summary

Fix stage auto-update not working when tickets are received via fetchmail,
and prevent a crash when a new email creates a ticket for the first time.

### 1. `TypeError` on new incoming emails

When a brand new email arrived with no existing ticket, `routes[0][1]`
(the ticket ID) was `None`. The code called `int(None)` without any guard,
crashing with:

```
TypeError: int() argument must be a string, a bytes-like object
or a real number, not 'NoneType'
```

This happened every time a customer sent a new support request by email,
making it impossible to receive tickets from the inbox.

### 2. Stage never changed when email arrived via fetchmail

The original code identified the sender by looking up the **route user**
(`routes[0][3]`) and comparing their partner with the ticket customer:

```python
partner_id = (
    self.env["res.users"]
    .search([("id", "=", routes[0][3])], limit=1)
    .partner_id.id
)
if partner_id == ticket.partner_id.id:
    ...
```

When emails are processed via fetchmail, that user is always **OdooBot or
the cron user**, never the actual customer. So the partner comparison always
failed and the stage update never happened — the feature was silently broken
for all email-based workflows.

### 3. `browse()` result not validated against the database

`browse(id)` in Odoo always returns a non-empty recordset even if the record
does not exist in the database. A bare `if ticket:` check would always be
`True` for any non-zero ID. `ticket.exists()` was added to perform the
actual database check.

### `models/mail_thread.py`

- Added `if ticket_id:` guard before `browse()` to safely skip new incoming
  emails with no existing ticket
- `message_dict` is now passed to `change_status_ticket_from_portal` so the
  real email sender (`author_id`) can be read from the parsed message headers
- Replaced the route user lookup with `message_dict.get("author_id")` to
  correctly identify the customer regardless of how the email was delivered
  (portal, fetchmail, etc.)
- Replaced `if ticket:` with `ticket.exists()` for a proper database
  existence check

**Before:**

```python
def change_status_ticket_from_portal(self, routes):
    if routes and routes[0][0] == "helpdesk.ticket":
        ticket_id = routes[0][1]
        ticket = self.env["helpdesk.ticket"].sudo().browse(int(ticket_id))
        partner_id = (
            self.env["res.users"]
            .search([("id", "=", routes[0][3])], limit=1)
            .partner_id.id
        )
        if (
            ticket
            and partner_id == ticket.partner_id.id
            and ticket.team_id.autoupdate_ticket_stage
            and ticket.stage_id in ticket.team_id.autopupdate_src_stage_ids
        ):
            ticket.stage_id = ticket.team_id.autopupdate_dest_stage_id.id
```

**After:**

```python
def change_status_ticket_from_portal(self, routes, message_dict=None):
    if routes and routes[0][0] == "helpdesk.ticket":
        ticket_id = routes[0][1]
        if ticket_id:
            ticket = self.env["helpdesk.ticket"].sudo().browse(int(ticket_id))
            author_id = message_dict and message_dict.get("author_id")
            if (
                ticket.exists()
                and author_id == ticket.partner_id.id
                and ticket.team_id.autoupdate_ticket_stage
                and ticket.stage_id in ticket.team_id.autopupdate_src_stage_ids
            ):
                ticket.stage_id = ticket.team_id.autopupdate_dest_stage_id.id
```

### `tests/test_partner_response.py`

- `message_process()` helper now accepts an optional `email_from` parameter
  to allow simulating different senders in tests
- Added `test_no_change_stage_non_customer_through_mail`: verifies that an
  email from a non-customer (e.g. internal user) does **not** trigger a stage
  change — this is the key behavioral test for the `author_id` fix
- Added `test_new_ticket_via_email_no_crash`: reproduces the `TypeError`
  scenario (no `thread_id`) and ensures no exception is raised
- Added `test_ticket_without_team_no_crash`: ensures a ticket with no team
  assigned does not crash on an incoming email reply; stage must remain
  unchanged since there is no autoupdate configuration
- Added docstrings to all tests to document the intent of each scenario

## Test Coverage

| Test | Scenario |
|---|---|
| `test_change_stage_customer_answered` | Customer replies via portal → stage changes |
| `test_no_change_stage_customer_answered` | Source stage not in configured list → no change |
| `test_change_stage_deactivated` | Autoupdate disabled on team → no change |
| `test_change_stage_customer_answered_through_mail` | Customer replies via fetchmail → stage changes |
| `test_no_change_stage_customer_answered_through_mail` | Source stage not in configured list → no change |
| `test_change_stage_deactivated_through_mail` | Autoupdate disabled → no change |
| `test_no_change_stage_non_customer_through_mail` | Email from non-customer → no change (**new**) |
| `test_new_ticket_via_email_no_crash` | New email, no existing ticket → no crash (**new**) |
| `test_ticket_without_team_no_crash` | Ticket with no team → no crash, stage unchanged (**new**) |

cc @agent-z28

### DIFF
--- a/helpdesk_ticket_partner_response/models/mail_thread.py
+++ b/helpdesk_ticket_partner_response/models/mail_thread.py
@@ -6,28 +6,27 @@ class MailThread(models.AbstractModel):
 
     @api.model
     def _message_route_process(self, message, message_dict, routes):
-        self.change_status_ticket_from_portal(routes)
+        self.change_status_ticket_from_portal(routes, message_dict)
         return super()._message_route_process(message, message_dict, routes)
 
-    def change_status_ticket_from_portal(self, routes):
+    def change_status_ticket_from_portal(self, routes, message_dict=None):
         if routes and routes[0][0] == "helpdesk.ticket":
             ticket_id = routes[0][1]
-            if ticket_id:
-                ticket = self.env["helpdesk.ticket"].sudo().browse(int(ticket_id))
-                if ticket and routes[0][3]:
-                    partner_id = (
-                        self.env["res.users"]
-                        .search([("id", "=", routes[0][3])], limit=1)
-                        .partner_id.id
-                    )
-                    if partner_id:
-                        if (
-                            ticket
-                            and partner_id == ticket.partner_id.id
-                            and ticket.team_id.autoupdate_ticket_stage
-                            and ticket.stage_id
-                            in ticket.team_id.autopupdate_src_stage_ids
-                        ):
-                            ticket.stage_id = (
-                                ticket.team_id.autopupdate_dest_stage_id.id
-                            )
+            # When the email creates the ticket/thread, Odoo can route the message
+            # with a missing (None) id. In that case we must not crash.
+            if ticket_id is None:
+                return
+            try:
+                ticket_id_int = int(ticket_id)
+            except (TypeError, ValueError):
+                return
+
+            ticket = self.env["helpdesk.ticket"].sudo().browse(ticket_id_int)
+            author_id = message_dict and message_dict.get("author_id")
+            if (
+                ticket.exists()
+                and author_id == ticket.partner_id.id
+                and ticket.team_id.autoupdate_ticket_stage
+                and ticket.stage_id in ticket.team_id.autopupdate_src_stage_ids
+            ):
+                ticket.stage_id = ticket.team_id.autopupdate_dest_stage_id.id
--- a/helpdesk_ticket_partner_response/tests/test_partner_response.py
+++ b/helpdesk_ticket_partner_response/tests/test_partner_response.py
@@ -76,12 +76,12 @@ class TestCustomerResponse(HttpCaseWithUserPortal):
         )
         return ticket
 
-    def message_process(self):
+    def message_process(self, email_from=None):
         MailThread = self.env["mail.thread"]
         message = MAIL_TEMPLATE.format(
             to=self.env.user.email,
             subject="Your ticket has been created !!",
-            email_from=self.partner_portal.email,
+            email_from=email_from or self.partner_portal.email,
             msg_id="168242744424.20.2028152230359369389@dd607af32154",
         )
         MailThread.message_process(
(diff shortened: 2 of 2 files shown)


peluko00 (approved):
LGTM! Code review

pedrobaeza:
The issue is mentioning a problem in the previous line:

```python
ticket = self.env["helpdesk.ticket"].sudo().browse(int(ticket_id))
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

so this is not solving that problem, as the line remains untouched.

CristianoMafraJunior:
> The issue is mentioning a problem in the previous line:
> 
> ```python
> ticket = self.env["helpdesk.ticket"].sudo().browse(int(ticket_id))
> TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
> ```
> 
> so this is not solving that problem, as the line remains untouched.

The email routing can send `routes[0][1]` as `None` **(when the ticket/thread is being created)**, so the old code crashed at `int(ticket_id)`.
I changed `change_status_ticket_from_portal()` to guard against that case:
- If `ticket_id` is `None` (or cannot be converted to an integer), the method simply returns without crashing.
- Only when we have a valid numeric `ticket_id` do we `browse()` the ticket and apply the stage auto-update logic.

helpdesk_ticket_partner_response/models/mail_thread.py
```diff
@@ -6,28 +6,27 @@ class MailThread(models.AbstractModel):
...
-                            and ticket.team_id.autoupdate_ticket_stage
-                            and ticket.stage_id
-                            in ticket.team_id.autopupdate_src_stage_ids
-                        ):
-                            ticket.stage_id = (
-                                ticket.team_id.autopupdate_dest_stage_id.id
-                            )
+            # When the email creates the ticket/thread, Odoo can route the message
+            # with a missing (None) id. In that case we must not crash.
+            if ticket_id is None:
```
pedrobaeza, line 17: No need of this one.

helpdesk_ticket_partner_response/models/mail_thread.py
```diff
@@ -6,28 +6,27 @@ class MailThread(models.AbstractModel):
...
+                return
+            try:
+                ticket_id_int = int(ticket_id)
+            except (TypeError, ValueError):
+                return
+
```
pedrobaeza, line 23: Don't add empty lines inside the method.

marcelsavegnago:
ping @CristianoMafraJunior 

image

PR #875, opened 2025-10-30
[16.0][ADD] helpdesk_mgmt_stock

rousseldenis:

PR #965, opened 2026-03-12
[16.0][ADD] helpdesk_mgmt_stock_origin: pre-fill heldesk tickets created from a picking using origin PO/SO

nicolas-delbovier-acsone:
Depends on https://github.com/OCA/helpdesk/pull/875


helpdesk_mgmt_stock_origin/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "summary": """Pre-fill the sale order or purchase order fields of helpdesk
+    tickets created from a picking based on the origin field of the picking""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/helpdesk",
+    "depends": [
+        "helpdesk_mgmt_stock",
+        "helpdesk_mgmt_sale",
+        "helpdesk_mgmt_purchase",
```
lmignon, line 15: test dependencies are missing mgmt_sale and mgmt_purchase....
IMO this addon should be split into 2 addons. One for sale and one for purchase to avoid to depends of sale and purchase at same time into the same addon...

lmignon (changes requested):

helpdesk_mgmt_stock_origin/__manifest__.py
nicolas-delbovier-acsone: > test dependencies are missing mgmt_sale and mgmt_purchase....

My bad I believed I needed these dependencies first so I added them in manifest only to later realize that the code was not using those modules logic so I removed the test dep but not in the manifest

> IMO this addon should be split into 2 addons. One for sale and one for purchase to avoid to depends of sale and purchase at same time into the same addon...

Good idea, I will close this PR and create 2 new ones

nicolas-delbovier-acsone:
See #968 and #969

PR #966, opened 2026-03-12
[18.0][FIX] helpdesk_mgmt_project: Remove store from counts

christian-ramos-tecnativa:
Having count fields that get computed on the creation of a different record can lead into a concurrence error rolling back the records created, in this case the helpdesk tickets. Didn't find any other reason for the stores than the domain in the views so I adapt them to a similar one.

@Tecnativa TT61446
@pedrobaeza @victoralmau

### DIFF
--- a/helpdesk_mgmt_project/models/project.py
+++ b/helpdesk_mgmt_project/models/project.py
@@ -7,7 +7,7 @@ class ProjectProject(models.Model):
     ticket_ids = fields.One2many(
         comodel_name="helpdesk.ticket", inverse_name="project_id", string="Tickets"
     )
-    ticket_count = fields.Integer(compute="_compute_ticket_count", store=True)
+    ticket_count = fields.Integer(compute="_compute_ticket_count")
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
Please ping the original author of that store change to know their reasons.

christian-ramos-tecnativa:
Hello @DarioLodeiros, do you remember by any chance if there was any additional reason to make those counters stored, beyond enabling the “Open tickets” filter in https://github.com/OCA/helpdesk/commit/3757263fddf1c026633cf281616159e20ac9e5b7?

pedrobaeza:
That filter “Open tickets” can be done adding a `search` method on the field.

christian-ramos-tecnativa:
Don't you think the way I changed the domain will also provide the same result?

pedrobaeza:
Oh, yeah, creative way!

luisDIXMIT (approved):
Code review, LGTM!

PR #967, opened 2026-03-12
Draft: Split timesheet control

Hugo-Trentesaux:
Splits module `helpdesk_mgmt_timesheet` to remove `project_timesheet_time_control` dependency.

This adds a new module `helpdesk_mgmt_time_control`.

I tested with and without, but I still have to:
- fix translation files (any help to automate that?)
- fix documentation
- run tests and pre-commit


Hugo-Trentesaux:
Not stale, I just have to find some more time to finish it.

PR #974, opened 2026-04-17
[18.0][IMP] helpdesk_mgmt: improve portal ticket list UI and default groupby by stage

JulienMartinez:
## What

- Visual overhaul of the ticket list for a cleaner and more concise layout
- Add thicker bottom border on table header for better visual separation
- Display closed tickets in italic and muted style
- Set stage as default groupby in portal ticket list

### Before
Screenshot_20260417_121944

### After
Screenshot 2026-04-17 at 12-27-22 Odoo


pedrobaeza:
Can you put screenshots of before/after?

JulienMartinez:
> Can you put screenshots of before/after?

@pedrobaeza  Hi, I added before/after screenshots

pedrobaeza:
Thanks for the screenshots. I don't like to group by stage, but to keep create date order, as some newer tickets can be in a more advanced stage, and this way, you don't see chronologically.

About the rest of the changes, I consider that the assigned user is not relevant in the list, but the closed date is.

You have to check also the CI, and rebase instead of doing `git pull`.

DorianMAG (changes requested):
Hi,
thx for the proposition.
Can we keep grouby to None by default, and remove user_id in view list plz?

flotho:
> Thanks for the screenshots. I don't like to group by stage, but to keep create date order, as some newer tickets can be in a more advanced stage, and this way, you don't see chronologically.
> 
> About the rest of the changes, I consider that the assigned user is not relevant in the list, but the closed date is.
> 
> You have to check also the CI, and rebase instead of doing `git pull`.

Thanks @pedrobaeza , 

I totally understand that having the creation date could be important , especially if you have some SLA delay for answering.

Yet I would like to share with you that this design has been asked by some of our customers who are following frequently the work in progress on our portal. They're focused on the resolution not on the creation date, in an 'agile' way.

Thus, they're considering that those fields are not relevant in a list view : 
* The By field because they're checking tickets for their companies
* Among 3 date fields, the last answer seems to be the most important

The user_id is merely a matter of taste from our part.

So the question is, how can we reconcile those 2 PoV : 
* Not grouping by default is easy
* Hiding fields should be agreed

Anyway, thanks for your feedback

pedrobaeza:
If not agreed like in this case, I think you should put on your custom the things you want to have, like hiding or adding columns. I do that on my instance for "debatable things".

flotho:
> If not agreed like in this case, I think you should put on your custom the things you want to have, like hiding or adding columns. I do that on my instance for "debatable things".

Thanks, we won't push on this one. It'll return to our own scope.

Regards

PR #985, opened 2026-04-28
[17.0][IMP] helpdesk_mgmt: Add default fields to partner_id field

victoralmau:
Add default fields to partner_id field

Similar to what happens with CRM, the following changes are added to the partner_id field:
- Define a context with the default values for the fields that exist in the ticket

Please @pedrobaeza and @Andrii9090-tecnativa can you review it?

@Tecnativa TT62128

### DIFF
--- a/helpdesk_mgmt/views/helpdesk_ticket_views.xml
+++ b/helpdesk_mgmt/views/helpdesk_ticket_views.xml
@@ -195,7 +195,10 @@
                             <field name="channel_id" />
                         </group>
                         <group>
-                            <field name="partner_id" />
+                            <field
+                                name="partner_id"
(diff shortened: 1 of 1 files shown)


Andrii9090-tecnativa (approved):

helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -195,7 +195,11 @@
                             <field name="channel_id" />
                         </group>
                         <group>
-                            <field name="partner_id" />
+                            <field
+                                name="partner_id"
+                                widget="res_partner_many2one"
```
pedrobaeza: I don't think you should add this, that it's something of `partner_autocomplete` module.

```suggestion
```

pedrobaeza (approved):
/ocabot merge patch

PR #1000, opened 2026-05-11
[18.0][IMP] helpdesk_ticket_partner_response: stage change via email and portal

KDRdev:
Currently, sending an email or replying as non-logged in user via portal may not trigger stage change. This PR aims to fix this behaviour.


peluko00 (approved):
LGTM!

PR #1003, opened 2026-05-18
[18.0][ADD] helpdesk_ticket_message_contact_default_unchecked

IriaAlonso:
By default, when you create a ticket with a selected contact or send a message, the “Add as follower and recipient” checkbox is checked.
With this addon, the checkbox is unchecked by default.


IriaAlonso:
@flotho @JulienMartinez could you review it please?

helpdesk_ticket_message_contact_default_unchecked/static/tests/tours/TicketContactUncheck.tour.esm.js
```diff
@@ -0,0 +1,26 @@
+/* Copyright 2026 Solvos - Iria Alonso
```
etobella, line 1: Would be better to do a hoot test, isn't it?

etobella (commented):
A small suggestion

PR #978, opened 2026-04-21
[19.0] [MIG] helpdesk_mgmt: Migration to 19.0

stferraro:

PR #1006, opened 2026-05-25
[19.0] [MIG] helpdesk_mgmt_sla: Migration to 19.0

Tomero-Dev:


Tomero-Dev:
The module works completely, but because the base module helpdesk_mgmt is not validated in the main repository, it breaks in the github action of BUILD.
PR depends: https://github.com/OCA/helpdesk/pull/978

pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #979, opened 2026-04-27
[19.0][MIG] helpdesk_type: Migration to 19.0

meetscs:
Helpdesk Ticket Type
==================
This module adds a type field on the helpdesk ticket.

PR #1009, opened 2026-05-25
[19.0][MIG] helpdesk_type_sla:Migration to 19.0

Tomero-Dev:


Tomero-Dev:
The module works completely, but because the base module helpdesk_mgmt is not validated in the main repository, it breaks in the github action of BUILD.
PR depends: https://github.com/OCA/helpdesk/pull/979

PR #1017, opened 2026-05-27
[19.0][MIG] helpdesk_mgmt_fieldservice_equipment: Migration to 19.0

Tomero-Dev:


helpdesk_mgmt_fieldservice_equipment/__manifest__.py
```diff
@@ -3,12 +3,12 @@
...
     "name": "Helpdesk Ticket Field Service Equipment",
     "summary": "Helpdesk Ticket Field Service Equipment",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/helpdesk",
     "license": "AGPL-3",
     "category": "HelpDesk Service",
-    "depends": ["fieldservice", "helpdesk_mgmt_fieldservice"],
+    "depends": ["base", "fieldservice", "helpdesk_mgmt_fieldservice"],
```
BhaveshHeliconia: @Tomero-Dev Could you explain why you added `base` to the `depends` ?

Tomero-Dev: Thanks for the note. 
I added it when I got an initial error using the External ID of base.user_demo in the tests.

When I read about the changes in Odoo 19 regarding tests and that these references are no longer allowed, I changed everything, generating the test data in setUpClass.
That line remained there.

If you think it's necessary, I can remove it, although ultimately, base is always installed, but it's up to you.
Any advice is welcome.

BhaveshHeliconia: If there’s no specific logic requiring it, then I think we can remove it. Since `base` is already installed by default, keeping it in `depends` does not seem necessary in this case.

Tomero-Dev: I've made the changes as you requested.

Thank you for the feedback. 😉

helpdesk_mgmt_fieldservice_equipment/tests/test_fsm_equipment.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestFSMEquipment(TransactionCase):
```
BhaveshHeliconia, line 7:
```suggestion
class TestFSMEquipment(Basecommon):
```

helpdesk_mgmt_fieldservice_equipment/tests/test_fsm_equipment.py
```diff
@@ -0,0 +1,61 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Create test data
+        cls.location_partner_id = (
+            cls.env["res.partner"].create({"name": "Loc Partner 1"}).id
+        )
+
```
BhaveshHeliconia, line 15:
```suggestion
```

helpdesk_mgmt_fieldservice_equipment/tests/test_helpdesk_ticket.py
```diff
@@ -0,0 +1,123 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase, new_test_user
+
+from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
+
+
+class TestHelpdeskTicketEquipment(TransactionCase):
```
BhaveshHeliconia, line 10:
```suggestion
class TestHelpdeskTicketEquipment(Basecommon):
```

helpdesk_mgmt_fieldservice_equipment/tests/test_helpdesk_ticket.py
```diff
@@ -0,0 +1,123 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
```
BhaveshHeliconia, line 14:
```suggestion
```

helpdesk_mgmt_fieldservice_equipment/tests/test_helpdesk_ticket.py
```diff
@@ -0,0 +1,123 @@
...
+        # Create test data
+        cls.partner = cls.env["res.partner"].create({"name": "Partner 1"})
```
BhaveshHeliconia, line 16:
```suggestion
```

BhaveshHeliconia (commented):
Consider using BaseCommon instead of TransactionCase. In this case, there’s no need to redefine cls.env with tracking_disable=True, and res.partner is already available via BaseCommon. This would help simplify and optimize the test code.

helpdesk_mgmt_fieldservice_equipment/tests/test_fsm_equipment.py
Tomero-Dev: Okay, those tests come directly from version 18, so I haven't modified practically anything, only the essentials.
I'll implement the changes you suggested. 
Thanks.

PR #1020, opened 2026-06-02
[19.0][IMP] helpdesk_mgmt: team-based automatic ticket assignment

gerard-domatix:
## What does this PR do?

Adds configurable automatic assignment of new tickets to team members.

A new `assign_method` field on `helpdesk.ticket.team` controls the strategy:

- **Manually** (default): no automatic assignment.
- **Random (round-robin)**: cycles through team members in deterministic order.
- **Balanced (least busy)**: assigns to the member with fewest open tickets.
- **By Tags**: assigns to a member mapped to the ticket's tags (least busy
  among eligible); mapping configured via the new *Tag Assignments* tab on
  the team form.

The new `helpdesk.ticket.team.assignment.tag` model stores the tag→members
mapping. Tickets created in a folded/closed stage are never auto-assigned.
For non-manual teams, the company-level "assign to creator" default is
overridden so the team strategy takes precedence.

### DIFF
--- a/helpdesk_mgmt/README.rst
+++ b/helpdesk_mgmt/README.rst
@@ -103,6 +103,23 @@ Teams
 
 |image4|
 
+Automatic assignment
+~~~~~~~~~~~~~~~~~~~~
+
+Each team has an *Assignment Method* that decides how new unassigned
+tickets get an assigned user:
+
+- *Manually*: no automatic assignment (default).
+- *Random (round-robin)*: rotate cyclically over the team members.
+- *Balanced (least busy)*: assign to the member with the fewest open
+  tickets.
+- *By Tags*: assign to a member mapped to the ticket tags, picking the
+  least busy one. Fill in the *Tag Assignments* tab to map each tag to
+  its eligible members.
+
+Tickets created directly in a folded or closed stage are never
+auto-assigned.
+
 Tags
 ----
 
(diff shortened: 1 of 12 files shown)


ppyczko:
Hi @gerard-domatix thanks for the contribution! Looking at the proposed changes, I noticed we already have the [helpdesk_mgmt_assign_method](https://github.com/OCA/helpdesk/tree/17.0/helpdesk_mgmt_assign_method) module in the 17.0 branch, which sets up the foundation for ticket assignment logic.

Wouldn't it be better to migrate and reuse helpdesk_mgmt_assign_method instead of introducing these changes inside helpdesk_mgmt? This way, we keep the core module clean and lightweight, and we just need to add this new assign method via tags as a new selection option inside the assignment extension. What do you think?

gerard-domatix:
> Hi @gerard-domatix thanks for the contribution! Looking at the proposed changes, I noticed we already have the [helpdesk_mgmt_assign_method](https://github.com/OCA/helpdesk/tree/17.0/helpdesk_mgmt_assign_method) module in the 17.0 branch, which sets up the foundation for ticket assignment logic.
> 
> Wouldn't it be better to migrate and reuse helpdesk_mgmt_assign_method instead of introducing these changes inside helpdesk_mgmt? This way, we keep the core module clean and lightweight, and we just need to add this new assign method via tags as a new selection option inside the assignment extension. What do you think?

Good point!  I'll migrate helpdesk_mgmt_assign_method and keep helpdesk_mgmt clean. Thanks for the review.

PR #1022, opened 2026-06-03
[18.0][FIX] helpdesk_ticket_partner_response: compare to normalized ticket email

KDRdev:
Currently, if we receive response on ticket from customer with "from" address `"test@domain.com"` and ticket email is set as `"TEST_CUSTOMER <test@domain.com>"`, the stage won't get updated automatically. This fix aims to cover such case.


peluko00 (approved):
LGTM!

PR #1038, opened 2026-06-12
[18.0][MIG] helpdesk_mgmt_assign_method: Migration to 18.0

gerard-domatix:


ppyczko:
Hi @gerard-domatix could you please preserve the original commit history? The OCA migration guide explains the expected workflow in the "How-to" section, where the original commits are replayed and only the migration-specific changes go into the final MIG commit.

For reference, here's an example where the module history is preserved:

[Migration guide ("How-to")](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#how-to)
[Example commit history (helpdesk_mgmt_assign_method)](https://github.com/OCA/helpdesk/commits/17.0/helpdesk_mgmt_assign_method)

PR #1046, opened 2026-06-15
[18.0][FIX] helpdesk_mgmt: Customer must not receive an internal notification

ypapouin:
Fix longstanding bug where stage changes are possibly sent to external subscribers.

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -314,7 +314,7 @@ class HelpdeskTicket(models.Model):
                     "composition_mode": "mass_mail",
                     "auto_delete_keep_log": False,
                     "subtype_id": self.env["ir.model.data"]._xmlid_to_res_id(
-                        "mail.mt_note"
+                        "mail.mt_comment"
                     ),
                     "email_layout_xmlid": "mail.mail_notification_light",
                 },


SirPyTech (changes requested):
Hi, thanks for your PR!
I noticed this behavior too but I think we don't have to change this subtype to fix it, moreover it is normal for the customer to receive stage updates about the ticket for certain stages.

I believe the problem is that the module's email template includes the customer (`partner_id`) in the recipients: https://github.com/OCA/helpdesk/blob/12616f4c290b5de838e3574451cd9d4d1c4d5984/helpdesk_mgmt/data/helpdesk_data.xml#L142-L148
You can either change the template or create a new one that does not address the customer and assign it to the stage you need
Please let me know what you think

I noticed this PR because I am fixing a somewhat related issue in https://github.com/OCA/helpdesk/pull/1090, please have a look if you have some time.
Have a good day 🌞

ypapouin:
@SirPyTech, changing the type should not change anything to the behaviour, the partner will continue to receive a notification. The change is only to be consistent with Odoo codebase: a `mail.mt_note` should never reach a partner since a note is considered private.

SirPyTech (approved):
> @SirPyTech, changing the type should not change anything to the behaviour, the partner will continue to receive a notification. The change is only to be consistent with Odoo codebase: a `mail.mt_note` should never reach a partner since a note is considered private.

Sorry I misunderstood this PR's purpose!
So the purpose is simply to use an external subtype because the generated email will be sent to the customer, that seems right 👍
I reviewed the code and tried it locally and can confirm that the behavior does not change.

PR #1049, opened 2026-06-19
[18.0][IMP] helpdesk_mgmt exclude emails from auto reply

jans23:
This is particularly useful to prevent auto-reply ping pongs between Odoo and other ticketing systems.


jans23:
#1050 and #1051 are better approaches to tackle this problem. Closing...

PR #1050, opened 2026-06-22
[18.0][ADD] helpdesk_mgmt_mail_autogenerated_header

jans23:
When a helpdesk ticket is created from an auto-generated email (e.g. an
out-of-office reply carrying `Auto-Submitted: auto-replied` or
`X-Auto-Response-Suppress: AutoReply`/`All`), the stage welcome email
configured on the ticket stage must **not** be sent back to avoid an email loop.

This module integrates `helpdesk_mgmt` with `mail_autogenerated_header`:
the latter detects such headers on incoming messages and sets a context key
that this module reads in `_track_template_skip_autoreply` to suppress the
outgoing welcome email.

If follows a similar approach as #1051.

PR #1054, opened 2026-06-25
[18.0][IMP] helpdesk_mgmt to ignore auto replies

jans23:
Together with #1050 this avoids sending automatic welcome email as result of other automatic email being received.


jans23:
@victoralmau could you review this, please?

victoralmau (changes requested):
In my opinion, this should be done in `helpdesk_mgmt_mail_autogenerated_header` (https://github.com/OCA/helpdesk/pull/1050) by overriding the `_track_template()` method and removing `res[“stage_id”]` when appropriate (depending on what the `_track_template_skip_autoreply()` method returns).

PR #1055, opened 2026-07-02
[18.0][FIX] helpdesk_mgmt: do not reset stage_id when user_id changes within the same team

JulienMartinez:
## Problem

`_compute_stage_id` unconditionally resets `stage_id` to the first
applicable stage of the team whenever it is triggered, even when the
current stage is already valid for that team.

Since `stage_id` depends on `team_id`, and `team_id` depends on
`user_id`, changing `user_id` on a ticket marks `team_id` as needing
recomputation, which in turn marks `stage_id` as needing
recomputation as well - even if `team_id` doesn't actually change
(e.g. reassigning a ticket to another agent within the same team).

As a result, simply reassigning a ticket to a different agent in the
same team resets its stage back to the first one, losing the actual
progress of the ticket.

## Solution

Only reset `stage_id` when the current stage is not part of the
applicable stages for the (possibly new) team. This preserves the
existing stage whenever it's still valid, while keeping the original
behavior when the team actually changes to one where the current
stage doesn't apply.

Verified `_get_applicable_stages()` correctly includes both
team-specific and shared/no-team stages, so this works consistently
across single-team, multi-team, and unassigned stage setups.

## Test

Added `test_stage_id_not_reset_on_user_change`, which fails without
the fix (stage reset to the first one) and passes with it.

### DIFF
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -19,8 +19,13 @@ class HelpdeskTicket(models.Model):
 
     @api.depends("team_id")
     def _compute_stage_id(self):
+        # This compute is executed on user change, even if not changing team, so let's
+        # apply a preventive check for not changing stage if the current one is still
+        # applicable to the current team
         for ticket in self:
-            ticket.stage_id = ticket.team_id._get_applicable_stages()[:1]
+            applicable_stages = ticket.team_id._get_applicable_stages()
+            if ticket.stage_id not in applicable_stages:
+                ticket.stage_id = applicable_stages[:1]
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please cherry-pick #814 instead.

JulienMartinez:
@pedrobaeza Done. I replaced my hand-written commit with a cherry-pick of the original fix from #814 , so authorship and commit message are preserved.

DorianMAG (approved):
Thx for this fix.
Code review and functional test.
LGTM

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Finishing the fw-port chain in #1056

PR #1062, opened 2026-07-16
[18.0][FIX] helpdesk_mgmt: preserve ticket stage on team change

jans23:
When a ticket was re-assigned to another team, the stage was reset back to the first applicable stage (e.g. 'New'), which re-triggered stage-based automatic emails. Now the current stage is kept if it is also applicable in the newly assigned team.


pedrobaeza:
Already done in #1055

PR #837, opened 2025-09-14
[IMP] helpdesk_mgmt_sla: Improve the way to handle it

etobella:
A lot of changes were done or need to be done:

- [x] creation of a new module for handling the SLAs per ticket
- [x] removal of cron
- [x] Modify security
- [x] migration scripts
- [x] Review affected OCA modules
- [x] Allow a way to handle SLA modfications (when a ticket doesn't anymore fit the SLAs)
- [x] Update README
- [x] Add SLA reports

A small view of how it looks:

image

@marcelsavegnago I am sure you will like it :wink:

PR #1063, opened 2026-07-16
[17.0][IMP] helpdesk_mgmt_sla: Improve the way to handle it

david-banon-tecnativa:
Backport of https://github.com/OCA/helpdesk/pull/837

Please review @etobella @christian-ramos-tecnativa
TT63668 @Tecnativa


etobella:
For me it is not a problem, however we might need to change migration scripts on 18 to avoid problems (because the migration could already be done on 17)

etobella (approved):

eduezerouali-tecnativa (approved):

pedrobaeza:
/ocabot merge nobump

Please change v18 migration scripts to detect if the conversion has already been done for not repeating it.

PR #1052, opened 2026-06-23
[18.0][MIG] website_helpdesk_mgmt: Migration from 13.0 to 18.0

jans23:
Note that this module doesn't allow using the field partner_email because it was like this in the original module for 13.0 and because I think it would allow email spoofing. Let me know if you think it should be white-listed.

PR #1065, opened 2026-07-16
[19.0][ADD] helpdesk_mgmt_portal_form

arielbarreiros96:
This module is intended to extend portal capability to create tickets, allowing to create tickets from a configurable form, very similar to what surveys do, but only on the front, since the results are stored directly on the ticket description. The idea is to prepare forms for each category, that the user can fill, and the final result on the ticket description is no longer user "random" text, but instead a structured HTML containing only relevant data. It even supports conditional questions, which is quite handy

I'm adding some screenshots, I think this version is stable enough for being v1
Captura de pantalla 2026-07-16 a las 18 12 22
Captura de pantalla 2026-07-16 a las 18 19 19
Captura de pantalla 2026-07-16 a las 18 12 13


pedrobaeza:
Hi, Ariel, instead having a specific development for this, I think you should simply expand website form to include the option "Create a ticket":

image

and then, you can customize with no more effort the fields:

image

In fact, it's already implemented in https://github.com/OCA/helpdesk/pull/1052

arielbarreiros96:
> Hi, Ariel, instead having a specific development for this, I think you should simply expand website form to include the option "Create a ticket":
> 
> image
> and then, you can customize with no more effort the fields:
> 
> image
> In fact, it's already implemented in #1052

Hello @pedrobaeza thanks for this insight, actually I think I could also use website_helpdesk_mgmt for the use case I'm trying to cover. But website_helpdesk_mgmt and helpdesk_mgmt_portal_form have different purposes even though the result is the same.

The way I see it website_helpdesk_mgmt is for creating a website form, and creating a ticket from that form, like with the contacts, leads, etc., which is very cool by the way, and it covers the case when the user is hasn't signed in on the portal for example. 

On the other hand helpdesk_mgmt_portal_form lives on the implementation of the helpdesk_mgmt portal form itself, so it uses the standard buttons for "+NEW" ticket which you only get to see from the portal after signing in, and despite being a full form, it writes everything only to the ticket description, since its purpose is only to gather the description on a structured way. 

On the other hand website_helpdesk_mgmt would allow me to map everything 1:1 at field level, like 1 form field=1 ticket field, but in this case is not what I pursued, since I just needed to fill the ticket description with structured data.

I hope all this makes sense 😅, I'm open to discussion anyways, and your opinion is always welcomed 🙂

pedrobaeza:
OK, it seems very specific, but if it's your need, go ahead with it.

PR #1008, opened 2026-05-25
[18.0][MIG] helpdesk_mgmt_stock: Migration to 18.0

david-banon-tecnativa:
A migration from [this](https://github.com/OCA/helpdesk/pull/875) PR to 18.0, changes:

- Removed `helpdesk_motive` dependency
- Migration to v18.

cc @Tecnativa
ping @rousseldenis

PR #1069, opened 2026-07-27
[18.0][ADD] helpdesk_mgmt_stock_motive: Add module to glue helpdesk_mgmt_stock and helpdesk_motive

david-banon-tecnativa:
As talked about in https://github.com/OCA/helpdesk/pull/1008#discussion_r3645374364.
This module was part of helpdesk_mgmt_stock in v16, but as we don't need heldpesk_motive if we only want helpdesk_mgmt and stock, this has been separated into a new module.
 @rousseldenis 
 @tecnativa


pedrobaeza:
Why not put it in the same PR?

david-banon-tecnativa:
Isn't part of the OCA guidelines to have only one module per PR?

pedrobaeza:
That's when both modules are in the previous version, but in this case, you are splitting one module into 2, so both should go in the same migration at the same time (although separate commits).

PR #1081, opened 2026-08-07
[18.0][FIX] helpdesk_mgmt_timesheet: Prevent an error if the user does not have access to the project

victoralmau:
Prevent an error if the user does not have access to the project

Use case example:
- Create a user with permissions under Helpdesk and Project > User
- Create a project with visibility set to Invited internal users (private)
- Create a ticket linked to the previously created user and project
- When attempting to access the ticket with the created user, the user should be able to access it successfully without any errors being displayed

Please @pilarvargas-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT63960


pilarvargas-tecnativa (approved):

christian-ramos-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1086, opened 2026-08-15
[FIX] helpdesk_mgmt: do not auto-assign tickets to portal users

nericervin:
## Problem

When *Auto assign User* is enabled on the company, a ticket submitted by a
portal user ends up assigned to that portal user. The customer who filed the
ticket becomes its assignee, which also means the ticket counts as assigned
and never shows up as unassigned, so nobody picks it up.

## How to reproduce

1. Enable *Auto assign User* on the company.
2. Reproduce what a portal controller does:

```python
env["helpdesk.ticket"].with_user(portal_user).sudo().create({
    "name": "Portal ticket",
    "description": "<p>...</p>",
    "team_id": team.id,
})
```

The ticket is created with `user_id` set to the portal user.

Note that `sudo()` does not change `env.uid`, it only elevates privileges, so
`self.env.user` inside `default_get()` is still the portal user.

## Cause

`helpdesk_mgmt/models/helpdesk_ticket.py`, `default_get()`:

```python
if "user_id" in fields and not defaults.get("user_id"):
    company = self.env["res.company"].browse(company_id)
    if company.helpdesk_mgmt_ticket_auto_assign:
        if defaults.get("team_id"):
            team = ...
            if self.env.user in team.user_ids:
                defaults["user_id"] = self.env.user.id
        else:
            defaults["user_id"] = self.env.user.id
```

The branch with a team is safe by accident: a portal user is not a member of
the team, so the membership check rejects it. The branch without a team
assigns unconditionally.

Portal controllers hit that second branch, because they pass the team in the
*values* of `create()` and not in the defaults: `team_id` is therefore not a
missing field, `default_get()` is not called for it, and
`defaults.get("team_id")` is falsy.

`helpdesk_mgmt`'s own portal controller happens to escape the issue, because
`_prepare_submit_ticket_vals()` sets `"user_id": False` explicitly. Any other
controller that does not set it — a custom website form, an API import — hits
the bug. A website form with `auth="public"` runs as `base.public_user`, so
in that case tickets end up assigned to the public user, which is even less
useful than being assigned to the customer.

## Fix

Skip the auto-assignment when the current user is a shared user
(`share == True`, i.e. portal or public). This is consistent with the domain
of `user_id` itself, which already excludes shared users:

```python
domain="team_id and [('share', '=', False),('id', 'in', user_ids)] or [('share', '=', False)]"
```

Internal users are unaffected: the existing behaviour (creator becomes the
assignee) is preserved, and a regression test covers it.

## Tests

Four tests added to `helpdesk_mgmt/tests/test_helpdesk_portal.py`, reusing the
existing `TestHelpdeskPortalBase` fixtures: portal user with and without a
team, `base.public_user`, and an internal user to make sure the current
behaviour does not change.

## Other branches

The same code is present in `17.0`. `16.0` is not affected: the
`helpdesk_mgmt_ticket_auto_assign` field does not exist there. Happy to
backport to `17.0` if you want it in the same PR or in a separate one.

### DIFF
--- a/helpdesk_mgmt/tests/test_helpdesk_portal.py
+++ b/helpdesk_mgmt/tests/test_helpdesk_portal.py
@@ -185,6 +185,22 @@ class TestHelpdeskPortal(TestHelpdeskPortalBase):
         self.assertEqual(resp.status_code, 200)
         self.assertIn("ticket-user-2", resp.text)
 
+    def test_portal_user_is_not_auto_assigned(self):
+        """A ticket created by a portal user is not assigned to them."""
+        # Not submitted through our own controller on purpose: it passes
+        # "user_id": False explicitly, so it never reaches the auto-assign
+        # default. This covers any other creator that leaves user_id out,
+        # such as a custom website form or an API import.
+        self.env.company.helpdesk_mgmt_ticket_auto_assign = True
+        # sudo() keeps env.uid: this is what portal controllers do.
+        ticket = (
+            self.env["helpdesk.ticket"]
+            .with_user(self.user_portal)
+            .sudo()
+            .create({"name": "Ticket", "description": "Description"})
+        )
+        self.assertFalse(ticket.user_id)
+
     def _count_close_buttons(self, resp) -> int:
         """Count close buttons in a form by counting forms with that target."""
         return resp.text.count('action="/ticket/close"')
--- a/helpdesk_mgmt/models/helpdesk_ticket.py
+++ b/helpdesk_mgmt/models/helpdesk_ticket.py
@@ -203,7 +203,14 @@ class HelpdeskTicket(models.Model):
         # If the team is set, the user must belong to that team.
         defaults = super().default_get(fields)
         company_id = defaults.get("company_id") or self.env.company.id
-        if "user_id" in fields and not defaults.get("user_id"):
+        # A shared user (portal, public) is never a valid assignee: the
+        # user_id domain already excludes them, and a portal customer must
+        # not end up assigned to the ticket they just submitted.
+        if (
+            "user_id" in fields
+            and not defaults.get("user_id")
+            and not self.env.user.share
+        ):
             company = self.env["res.company"].browse(company_id)
             if company.helpdesk_mgmt_ticket_auto_assign:
                 if defaults.get("team_id"):


helpdesk_mgmt/__manifest__.py
```diff
@@ -4,7 +4,7 @@
     "name": "Helpdesk Management",
     "summary": """
         Helpdesk""",
-    "version": "18.0.1.17.1",
+    "version": "18.0.1.17.2",
```
pedrobaeza: Don't modify module version. It will be done by the bot.

pedrobaeza (changes requested):
Please don't include slop tests generated by AI. You just need to check the new case, not all the variants.

nericervin:
Fair point. Reduced to a single test for the new case.

helpdesk_mgmt/tests/test_helpdesk_portal.py
```diff
@@ -204,3 +204,17 @@ def _call_close_ticket(self, ticket, stage):
         self.assertTrue(resp.is_redirect)  # http://127.0.0.1:8069/my/ticket/<ticket-id>
         self.assertTrue(resp.headers["Location"].endswith(f"/my/ticket/{ticket.id}"))
         return resp
+
+
+@tagged("post_install", "-at_install")
+class TestHelpdeskPortalAutoAssign(TestHelpdeskPortalBase):
```
SirPyTech: **question**: Why a new class? This way the `setupClass` is executed again

helpdesk_mgmt/tests/test_helpdesk_portal.py
```diff
@@ -204,3 +204,17 @@ def _call_close_ticket(self, ticket, stage):
...
+    def test_portal_user_is_not_auto_assigned(self):
+        self.env.company.helpdesk_mgmt_ticket_auto_assign = True
+        # sudo() keeps env.uid: this is what portal controllers do.
```
SirPyTech: **suggestion**: Instead of copying "what controllers do", please call the controller like it's done in the other tests (`self._submit_ticket`) to avoid code duplication.

helpdesk_mgmt/__manifest__.py
SirPyTech: **note**: You can skip updating the version because it will be automatically done by a bot when the PR is merged.

SirPyTech (changes requested):
Thanks for the PR!
I only reviewed the code, and it is mostly good 👏
The only part worrying me is duplicating the controller code in the test instead of calling the controller, please let me know what you think

helpdesk_mgmt/__manifest__.py
nericervin: Reverted, the version is back to 18.0.1.17.1. Sorry for the noise.

nericervin: Done, reverted in the last push.

helpdesk_mgmt/tests/test_helpdesk_portal.py
nericervin: You are right, no reason for a separate class. Moved the test into TestHelpdeskPortal.

helpdesk_mgmt/tests/test_helpdesk_portal.py
nericervin: I tried that first, but going through the controller makes the test pass with or without the fix.

`_prepare_submit_ticket_vals()` sets `"user_id": False` explicitly (controllers/main.py), so `user_id` is not a missing field and `default_get()` never runs for it. That path never reaches the auto-assign branch. In fact `test_submit_ticket_02` already submits through the controller as a portal user and asserts `assertFalse(new_ticket.user_id)`, and it is green on 18.0 today, before this patch.

The bug shows up with any other creator that leaves `user_id` out of the values: a custom website form, an API import, a plain `create()`. That is why the test creates the ticket directly instead of reusing `_submit_ticket`.

I added a comment on the test saying exactly this, so the next reader does not try to simplify it into the controller call. Happy to reword it if you prefer.

helpdesk_mgmt/tests/test_helpdesk_portal.py
SirPyTech: :+1:

helpdesk_mgmt/tests/test_helpdesk_portal.py
SirPyTech: Thanks for clarifying this!
I misunderstood the bugged use case because in the PR description you wrote
> a ticket submitted by a portal user ends up assigned to that portal user

reading this, I thought the bug was happening when a portal user submits a ticket using the UI/controller.

Adding a comment with a bit of explanation is the right move when the PR needs clarification :clap:
:+1:

helpdesk_mgmt/__manifest__.py
SirPyTech: :+1:

SirPyTech (approved):
Thanks for all the changes!
I only reviewed the code and it looks good.

If you have time, please have a look at the PRs I have opened a few days ago in this repo 🙏

PR #1088, opened 2026-08-17
[18.0][REF] helpdesk_mgmt: Simplify templates

SirPyTech:
Start simple with the email templates, they can be customized by the user anyway.
The new structure is aligned with the structure of `assignment_email_template` of the same module: https://github.com/OCA/helpdesk/blob/12616f4c290b5de838e3574451cd9d4d1c4d5984/helpdesk_mgmt/data/helpdesk_data.xml#L26-L27
The templates are `noupdate` records so existing installations are not affected.

A similar simplification was proposed in https://github.com/OCA/helpdesk/pull/644, but the PR was doing multiple things at once so it didn't get merged.


HekkiMelody (approved):
Code review, LGTM

pedrobaeza (commented):
Will they use the standard layout that already adds the decoration with the company details, buttons, signature and so on?

SirPyTech:
> Will they use the standard layout that already adds the decoration with the company details, buttons, signature and so on?

Thanks for having a look!
I tried it locally (why is there no runboat here?) and this is how the sent email looks like for changed state: 
image
The other templates should be similar.

Before the PR the same email was 
image
with company information duplicated.

The company details are still there, but I don't see any button or signature even in the removed code, which ones do you mean?

pedrobaeza (approved):
The button is to access the ticket with the token, but it might need a portal user. I see the simplification as OK, so let's continue:

/ocabot merge patch

Please fw-port it to 19

PR #1091, opened 2026-08-24
[IMP] helpdesk_mgmt: Show Access button in stage changed email

SirPyTech:
**Steps**:
1. A portal user opens a ticket
2. In the backend, an operator moves the ticket to a stage where a mail template is set
3. An email is sent to the portal user (even without this change)

**Current behavior**:
There is no link to open the ticket in the email, more precisely there is a link that is not visible because it has no content.

**Expected behavior**:
The email shows a link to open the ticket.

**Additional notes**:
In order to implement this, I reverted https://github.com/OCA/helpdesk/pull/508 that used `mass_mail` composition mode in order to ensure an email is sent when a ticket is created (ref. https://github.com/OCA/helpdesk/pull/508#discussion_r1341046828).

I reverted it because the creation email should be sent as a creation notification and not as a "stage change" email by setting the email template in the "New" stage: it is indeed weird to receive an email "Stage changed to New" when you have just submitted a ticket.

Moreover, the `mass_mail` composition mode is not needed in any other module implementing the same functionality (like https://github.com/odoo/odoo/blob/19fa4584c81bd3c636f244298d40e7fc1125e6f3/addons/project/models/project_task.py#L1576-L1580) and has the side effect of making the link in the email disappear (because `record_name` is set to `False` in https://github.com/odoo/odoo/blob/19fa4584c81bd3c636f244298d40e7fc1125e6f3/addons/mail/wizard/mail_compose_message.py#L967).

Without the `mass_mail` composition mode the email was indeed no more sent upon creation, that's why I added another commit to use the `mail_notify_author` context that actually fixes that issue, a test, and a few notes in **Configuration** about how such behavior should be achieved.

Sorry for the long explanation, email and its subtypes is a complicated topic so I spent some time searching through `mail.thread` model for the correct way to implement all this.
@victoralmau and @pedrobaeza you were involved in the specific discussion about this (https://github.com/OCA/helpdesk/pull/508#discussion_r1341046828) in the reverted PR and sure know more than me this topic, please let me know if I missed something or more in general what you think about this :pray:


pedrobaeza (commented):
But is this subscribing to "Ticket created" to the author of the email?

SirPyTech:
> But is this subscribing to "Ticket created" to the author of the email?

Hi, thanks for having a look!
Yes because I thought that's the standard way to notify a user about creation of a ticket.
That alone was not enough though, I also had to add the `mail_notify_author` in the context of the controller's `create` otherwise the author is usually not notified.

HekkiMelody (approved):
Code review, LGTM

PR #921, opened 2026-01-13
[16.0][IMP] helpdesk_mgmt_fieldservice: review fsm_order_close_wizard security

SirPyTech:
Proposing again a fix that was already proposed for `14.0` in https://github.com/OCA/helpdesk/pull/593 and https://github.com/OCA/helpdesk/pull/404, hoping the third time's the charm :four_leaf_clover:
I also added a test :rocket:

---

This commit reviews the security rules around the fsm_order_close_wizard.

The "Complete" button on the fsm.order triggered an access error for fieldservice.group_fsm_user_own. The commit allows fieldservice.group_fsm_user_own to use the wizard as well. However, the wizard is now only shown if the user also has write permission on the specific ticket (so as to play nice with a variety of setups, including helpdesk_mgmt.group_helpdesk_user_own).

If the user has permission to write on the ticket, the wizard will be shown as before; otherwise it will simply be skipped, but the fsm.order will be closed as it should.

There's also some code cleanup on the wizard, such as removing the unused team_id field.

PR #593, opened 2024-06-17
[14.0][IMP]helpdesk_mgmt_fieldservice: review fsm_order_close_wizard security

renda-dev:
proposing again #404 

@OCA/helpdesk-maintainers 

---

This commit reviews the security rules around the fsm_order_close_wizard.

The "Complete" button on the fsm.order triggered an access error for fieldservice.group_fsm_user_own. The commit allows fieldservice.group_fsm_user_own to use the wizard as well. However, the wizard is now only shown if the user also has write permission on the specific ticket (so as to play nice with a variety of setups, including helpdesk_mgmt.group_helpdesk_user_own).

If the user has permission to write on the ticket, the wizard will be shown as before; otherwise it will simply be skipped, but the fsm.order will be closed as it should.

There's also some code cleanup on the wizard, such as removing the unused team_id field.

PR #1092, opened 2026-08-25
[IMP] helpdesk_mgmt_fieldservice: review fsm_order_close_wizard security

SirPyTech:
Proposing again a fix that was already proposed for `14.0` in https://github.com/OCA/helpdesk/pull/593 and https://github.com/OCA/helpdesk/pull/404, and for `16.0` in https://github.com/OCA/helpdesk/pull/921 (including a test :rocket:) hoping the 4th time's the charm :four_leaf_clover:

---

This commit reviews the security rules around the fsm_order_close_wizard.

The "Complete" button on the fsm.order triggered an access error for fieldservice.group_fsm_user_own. The commit allows fieldservice.group_fsm_user_own to use the wizard as well. However, the wizard is now only shown if the user also has write permission on the specific ticket (so as to play nice with a variety of setups, including helpdesk_mgmt.group_helpdesk_user_own).

If the user has permission to write on the ticket, the wizard will be shown as before; otherwise it will simply be skipped, but the fsm.order will be closed as it should.

There's also some code cleanup on the wizard, such as removing the unused team_id field.

### DIFF
--- a/helpdesk_mgmt_fieldservice/models/fsm_order.py
+++ b/helpdesk_mgmt_fieldservice/models/fsm_order.py
@@ -13,7 +13,8 @@ class FSMOrder(models.Model):
     def action_complete(self):
         res = super().action_complete()
         if (
-            not self.ticket_id.stage_id.closed
+            self.ticket_id.has_access("write")
+            and not self.ticket_id.stage_id.closed
             and self.ticket_id.fsm_order_ids
             and all(self.ticket_id.mapped("fsm_order_ids.stage_id.is_closed"))
         ):
@@ -24,7 +25,6 @@ class FSMOrder(models.Model):
                 "target": "new",
                 "context": {
                     "default_ticket_id": self.ticket_id.id,
-                    "default_team_id": self.ticket_id.team_id.id,
                     "default_resolution": self.resolution,
                 },
             }
--- a/helpdesk_mgmt_fieldservice/models/fsm_order.py
+++ b/helpdesk_mgmt_fieldservice/models/fsm_order.py
@@ -12,11 +12,12 @@ class FSMOrder(models.Model):
 
     def action_complete(self):
         res = super().action_complete()
+        tickets = self.ticket_id
         if (
-            self.ticket_id.has_access("write")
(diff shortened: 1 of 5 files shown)


helpdesk_mgmt_fieldservice/models/fsm_order.py
```diff
@@ -12,19 +12,20 @@ class FSMOrder(models.Model):
 
     def action_complete(self):
         res = super().action_complete()
+        tickets = self.ticket_id
```
HekkiMelody, line 15: **nitpick**: this should be `ticket` since it can only be one

HekkiMelody (approved):
Code review, LGTM

helpdesk_mgmt_fieldservice/models/fsm_order.py
SirPyTech: Thanks for the review!
In theory, `self` can be a recordset containing multiple records so `self.ticket_id` can contain multiple records as well  because there is no `self.ensure_one()` here or in `super`'s method https://github.com/OCA/field-service/blob/84a6a28093538d6d31fafda8961f1d197ca2f227/fieldservice/models/fsm_order.py#L406-L411:
```py
    def action_complete(self):
        return self.with_context(bypass_order_completed_stage=True).write(
            {
                "stage_id": self.env.ref("fieldservice.fsm_stage_completed").id,
            }
        )
```
And I think it is using `write` to be able to update multiple records at once (it was needed in older versions).

In practice, I agree this is usually called on a single record.

HekkiMelody: In that case, the line

```py
 "default_ticket_id": tickets.id,
```

could fail with a SingletonError if there is more than one ticket, no?

SirPyTech: Right, sorry I missed that one!
So this override is actually forcing the method `fsm.order.action_complete` to only work on a single record while its definition allows it to accept multiple records :thinking: that's not good but I think fixing it is beyond this PR's scope.

HekkiMelody (approved):

PR #1093, opened 2026-08-25
[17.0][ADD] helpdesk_mgmt_partner_dashboard: New Module

cgarcia-solvos:
Replace the Helpdesk team dashboard with one grouped by partner.

imagen


helpdesk_mgmt_partner_dashboard/views/helpdesk_dashboard_views.xml
```diff
@@ -0,0 +1,198 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
dalonsod: There are two different models with views here, please separate them (a view file for each model)

helpdesk_mgmt_partner_dashboard/views/helpdesk_dashboard_views.xml
```diff
@@ -0,0 +1,198 @@
...
+        <field name="view_mode">kanban,tree,form</field>
+        <field name="context">
+            {'default_partner_id': active_id}
+        </field>
+        <field name="domain">[('partner_id', '=', active_id)]</field>
+    </record>
+
+    <record id="helpdesk_ticket_partner_kanban_view" model="ir.ui.view">
+        <field name="name">helpdesk.ticket.partner.kanban</field>
+        <field name="model">res.partner</field>
```
dalonsod, line 4: In order to prevent that this view is used as normal `kanban` view from other `res.partner` actions, set a low priority.

helpdesk_mgmt_partner_dashboard/views/helpdesk_dashboard_views.xml
```diff
@@ -0,0 +1,198 @@
...
+        <field name="arch" type="xml">
+            <kanban class="oe_background_grey o_kanban_dashboard" create="0">
```
dalonsod, line 7: Is it possible to add partner avatar/image to this view?

helpdesk_mgmt_partner_dashboard/views/helpdesk_dashboard_views.xml
```diff
@@ -0,0 +1,198 @@
...
+                    </t>
+                </templates>
+            </kanban>
+        </field>
+    </record>
+
+    <record id="helpdesk_ticket_partner_tree_view" model="ir.ui.view">
+        <field name="name">helpdesk.ticket.partner.tree</field>
+        <field name="model">res.partner</field>
+        <field name="arch" type="xml">
```
dalonsod, line 136: Regarding view priority, the same as above

dalonsod (changes requested):
Only code review, see initial comments.

And also, grouping partners attending at their hierarchy could be a better option. Only partners with no parent should be displayed, and for a certain partner, tickets count should include their tickets and for their childrens as well.

helpdesk_mgmt_partner_dashboard/views/helpdesk_dashboard_views.xml
cgarcia-solvos: All done!

helpdesk_mgmt_partner_dashboard/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Helpdesk Partner Dashboard",
+    "summary": """
+        Replace the Helpdesk team dashboard with one grouped by customer""",
+    "version": "17.0.1.0.0",
+    "license": "AGPL-3",
+    "category": "After-Sales",
+    "author": "Odoo Community Association (OCA)",
```
CristianoMafraJunior: Its author is missing.

CristianoMafraJunior (changes requested):

helpdesk_mgmt_partner_dashboard/__manifest__.py
cgarcia-solvos: True, thanks!

helpdesk_mgmt_partner_dashboard/models/res_partner.py
```diff
@@ -0,0 +1,61 @@
...
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    todo_ticket_count = fields.Integer(
+        string="Number of tickets",
+        compute="_compute_todo_tickets",
+        store=True,
```
dalonsod: By default, compute method for stored fields is executed as sudo (`compute_sudo=True`); when the field is not stored user's permissions are taken in account, because sudo is not used.

Then, `store=True` here has an undesired effect: users see that there are some tickets to do, and when are trying to access them, they're gone (e.g. those tickets that belong to other users, for a user with limited access to its own tickets). For these count fields `store=True` should be avoided.

Please take a look and try to move this field back to `store=False`, in order to be applied again user's permissions.

dalonsod (changes requested):
Functional review, see comment

helpdesk_mgmt_partner_dashboard/models/res_partner.py
cgarcia-solvos: Done!

dalonsod (approved):
Search technique for `todo_ticket_count` is a little strange for me, but it seems to work, for the desired purpose is ok.

dalonsod:
@ChristianSantamaria @IriaAlonso could you review? Thanks!

ChristianSantamaria (approved):

IriaAlonso (approved):
LGTM 👍

PR #1019, opened 2026-06-01
[18.0][IMP] helpdesk_mgmt_sale: Add sale_order_ids field to helpdesk ticket form view

victoralmau:
Add `sale_order_ids` field to helpdesk ticket form view

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT62704

PR #1098, opened 2026-08-27
[18.0][IMP] helpdesk_mgmt_sale: Remove the unnecessary order linking wizard

victoralmau:
Remove the unnecessary order linking wizard

The wizard is unnecessary because it is currently too restrictive (https://github.com/OCA/helpdesk/pull/1019#pullrequestreview-4400851976) and because this can already be done (defining orders) directly in the ticket field itself.

Related to https://github.com/OCA/helpdesk/pull/1019#pullrequestreview-5008640563 + https://github.com/OCA/helpdesk/pull/1019#pullrequestreview-5008640563

Please @carlos-lopez-tecnativa can you review it?

@Tecnativa


CristianoMafraJunior (approved):

carlos-lopez-tecnativa (commented):
LGTM to unify the domain in both ways, but please take into account this comment: https://github.com/OCA/helpdesk/pull/1019#pullrequestreview-4400851976

My original comment was about understanding why you added the many2many field **as editable** if there is already a wizard to do this. Could you first clarify the reason for making the many2many field editable if the wizard already handles this?

Now, we can either add this domain to the many2many field (which makes more sense to me), or, based on @pedrobaeza's comment, allow sales orders to be selected freely without restricting them to the partner. In that case, the solution would be to remove the domain from the wizard.

victoralmau:
> LGTM to unify the domain in both ways, but please take into account this comment: [#1019 (review)](https://github.com/OCA/helpdesk/pull/1019#pullrequestreview-4400851976)
> 
> My original comment was about understanding why you added the many2many field **as editable** if there is already a wizard to do this. Could you first clarify the reason for making the many2many field editable if the wizard already handles this?
> 
> Now, we can either add this domain to the many2many field (which makes more sense to me), or, based on @pedrobaeza's comment, allow sales orders to be selected freely without restricting them to the partner. In that case, the solution would be to remove the domain from the wizard.

Okay, now I understand what you meant; so maybe the wizard isn't necessary if it's currently too restrictive, right? The field in the form view was meant to visually display the linked order(s) and allow them to be linked easily (without the need for or restriction of the wizard). What do you think?

carlos-lopez-tecnativa (commented):
For simplicity, this change makes more sense. Before, linking a sales order to a ticket required more steps; now, I can select the sales order directly from the ticket.

I would prefer to get the opinion of other users, especially those who use the module. From one day to the next, the button to open the wizard disappears, which could be confusing for existing users.

Additionally, does it make sense to allow canceled orders to be selected? Or could we add a domain to exclude canceled orders? What do you think?

pedrobaeza (approved):
Cancelled orders may be selectable because the ticket can lead to cancel the SO.

/ocabot merge major

PR #981, opened 2026-04-27
[19.0][MIG] helpdesk_mgmt_crm: Migration to 19.0

meetscs:
Link between Helpdesk and CRM
=============================
This module allows you to convert a helpdesk ticket into an opportunity,
transferring its messages, and keeping both linked.

PR #1100, opened 2026-08-31
[19.0][MIG] helpdesk_mgmt_crm: Migration to version 19.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/helpdesk/pull/981
TT64104
@Tecnativa @pedrobaeza @victoralmau @BhaveshHeliconia could you please review this?


BhaveshHeliconia (approved):
Code and functional review LGTM!

helpdesk_mgmt_crm/models/helpdesk_ticket.py
```diff
@@ -18,28 +18,18 @@ class HelpdeskTicket(models.Model):
...
         )
-        mapped_data = {t["ticket_id"][0]: t["ticket_id_count"] for t in lead_data}
+        mapped_data = {ticket.id: count for ticket, count in lead_data}
         for item in self:
             item.lead_count = mapped_data.get(item.id, 0)
 
     def action_open_leads(self):
-        result = self.env["ir.actions.act_window"]._for_xml_id(
-            "crm.crm_lead_action_pipeline"
+        action = self.lead_ids._get_records_action(name=self.env._("Opportunity(ies)"))
```
pedrobaeza, line 31: Why this new translation?

helpdesk_mgmt_crm/tests/test_helpdesk_mgmt_crm.py
```diff
@@ -29,68 +29,71 @@ def setUpClass(cls):
         cls.ticket = cls.env["helpdesk.ticket"].create(
             {
                 "name": "Test ticket",
-                "partner_id": cls.partner.id,
+                "partner_id": cls.user.partner_id.id,
```
pedrobaeza: I don't think assigning the partner of the user is a good one.

helpdesk_mgmt_crm/tests/test_helpdesk_mgmt_crm.py
```diff
@@ -29,68 +29,71 @@ def setUpClass(cls):
...
                 "description": "Test description",
             }
         )
 
     @users("sale-user")
     def test_action_lead_create(self):
-        self.ticket.message_subscribe(
-            partner_ids=self.ticket.partner_id.ids,
+        # Re-browse in sale-user env to avoid admin env context mismatch
+        ticket = self.ticket.with_env(self.env)
```
pedrobaeza: This sounds weird

helpdesk_mgmt_crm/tests/test_helpdesk_mgmt_crm.py
```diff
@@ -29,68 +29,71 @@ def setUpClass(cls):
...
+        team = self.team.with_env(self.env)
+
```
pedrobaeza: Remove empty lines inside methods.

helpdesk_mgmt_crm/tests/test_helpdesk_mgmt_crm.py
pedrobaeza: And anyway, don't provoke this diff and use `self.ticket` again if needed.

pedrobaeza (changes requested):

helpdesk_mgmt_crm/models/helpdesk_ticket.py
carlos-lopez-tecnativa: Previously, the action was triggered using `_for_xml_id`. In this code, it is necessary to handle whether there is a single record, in which case the form view is displayed directly, or multiple records, in which case the list view is displayed. In this way, the action name was taken from the action.

To simplify the code, I suggested using [_get_records_action](https://github.com/odoo/odoo/blob/89650a5f44b5835028dba57a9ff6fa30515bc5ec/odoo/addons/base/models/ir_ui_view.py#L3248), which handles which view to display based on the number of records. However, this method does not set a name or take it from the action, so it is necessary to set an explicit name. Without it, when the list view is displayed, `unnamed` is shown in the breadcrumb.
image

helpdesk_mgmt_crm/tests/test_helpdesk_mgmt_crm.py
carlos-lopez-tecnativa: Sorry, this change came from another PR, and I didn't take it into account. I updated the code and some assertions to use message_partner_ids instead of message_follower_ids.mapped("partner_id").

Additionally, since the test now inherits from `BaseCommon`, tracking and subscription are disabled. I re-enabled tracking for the test to ensure that it works correctly.

pedrobaeza (approved):
/ocabot merge nobump

victoralmau (approved):
Code review OK

pedrobaeza:
@sbidoul we have traced a problem with curl 7.88 and a change in GitHub that is provoking this problem. Is there the possibility to update `curl` to a newer version in the ocabot instance?

sbidoul:
Does the our bot use curl? Do you have a link explaining the issue?

pedrobaeza:
We have followed an initial AI answer:

```
The real cause. The image includes curl 7.88.1 with nghttp2, which incorrectly sends the POST body of /git-upload-pack over HTTP/2. GitHub interprets this as an invalid anonymous request and responds with a 401 www-authenticate: Basic realm="GitHub", and Git ends up asking for a username.
```

and check a workaround and a possible solution:

- Workaround: avoid HTTP/2 using these environment variables:

  ```
  GIT_CONFIG_COUNT: "1"
  GIT_CONFIG_KEY_0: http.version
  GIT_CONFIG_VALUE_0: HTTP/1.1
  ```
- A possible solution: update curl version.

Both have been tested and working in local, but the workaround seems to be a short-term one, with that protocol being deprecated.

pedrobaeza:
Ah, and yes, `git` is using `curl` behind.

sbidoul:
Okay, but in a fresh ubunt:24.04 image, `apt update && apt install git && git clone https://github.com/OCA/maintainer-tools` now asks for a user name. So I suppose GitHub will fix that quickly. I'll check again tomorrow morning.

sbidoul:
It works with an Ubuntu 26.04.

Discussion at https://github.com/orgs/community/discussions/206581

pedrobaeza:
@sbidoul it seems GitHub has patched the underlying problem (but take it with a grain of salt still): https://github.com/Tecnativa/doodba/pull/725

ocabot is the one not even responding now.

sbidoul:
It's not fixed, the bot did not start because building the bot triggered the bug.

I added `git config --global http.version HTTP/1.1` in the bot dockerfile.

Can you retry the merge?

pedrobaeza:
Yes, it fails again. The only viable solution for now seems the workaround...

/ocabot merge nobump

sbidoul:
:shrug:

pedrobaeza:
This is another different thing! It seems they are rate-limiting unauthenticated downloads. This seems a similar measure like Linux kernel did some days ago: for avoiding consuming a lot of resources due to AI scrappers, they have forbidden this.

PR #999, opened 2026-05-11
[19.0][MIG] helpdesk_mgmt_timesheet: Migration to 19.0

pablolopez-domatix:
Migration of helpdesk_mgmt_timesheet from 18.0 to 19.0.

Changes done:
- Bump version to 19.0.1.0.0
- Replace `project_timesheet_time_control` dependency by `hr_timesheet_time_control`
- Replace `groups_id` by `group_ids` in tests (res.users field rename in v19)
- Replace `<data>` by `<odoo>` in XML views
- Add `Command` import and use `Command.set`/`Command.link` in tests
- Replace `setUp` by `setUpClass` in tests

PR #1102, opened 2026-09-04
[19.0]]MIG] helpdesk_mgmt_timesheet: Migration to 19.0

etobella:
Takes #999 and adds a small fix to fix the migration

/ocabot migration helpdesk_mgmt_timesheet


etobella:
@victoralmau @pedrobaeza I took the last version of 18 as base and just added the migration commit from #999. After that a small fix was necessary.

victoralmau (changes requested):
Code review OK but tests fail

Demo data cannot be used in the tests at this time, so it will need to be created within the tests.

PR #1106, opened 2026-09-14
[18.0][FIX] helpdesk_mgmt: link & subscribe customer partner for unknown senders

jans23:
Without this fix automatic responses couldn't be sent to emails/tickets being created from senders without partner record.


helpdesk_mgmt/models/helpdesk_ticket.py
```diff
@@ -326,6 +326,18 @@ def _track_template(self, tracking):
             )
         return res
 
+    def _helpdesk_link_and_subscribe_partner(self):
```
victoralmau, line 329: I don't understand the need for this method or why it's called at the end of `message_new()` when that method itself calls `_mail_find_partner_from_emails()` to retrieve the partners and subscribe them.

victoralmau (changes requested):

helpdesk_mgmt/models/helpdesk_ticket.py
jans23: [Here](https://github.com/Nitrokey/odoo-helpdesk/blob/02531f2646b6b77a56a8678e63c5bb3b5330cae0/helpdesk_mgmt/models/helpdesk_ticket.py#L362) only To and CC fields are parsed which doesn't catch From when the partner doesn't exist yet. I created a separate method to use it also from website_helpdesk_mgmt which otherwise faces the same issue.

victoralmau: Perhaps we need to define something like this for the `partner_email` field: https://github.com/odoo/odoo/blob/434b4e95a6ceca92de21e137911c0b72d7ef7a07/addons/crm/models/crm_lead.py# L100 so that everything works automatically without having to add any extra methods.

jans23: I don't see how that would solve the issue if no partner record exists. Please elaborate.

PR #1110, opened 2026-09-21
[18.0][ADD] helpdesk_mgmt_fieldservice_equipment_button

marcos-mendez:
Smart button on the equipment form that opens the helpdesk tickets of the equipment.

Depends on OCA/field-service#1623, which provides the button box on the equipment form.


marcos-mendez:
@etobella it is ok for you like this? Should i request review from some specific

etobella:
Silly question, shouldn't this go directly here?

https://github.com/OCA/helpdesk/tree/18.0/helpdesk_mgmt_fieldservice_equipment

I mean, a module just for adding a button seems too much for me. Maybe @max3903 has a bigger understanding, because I don't use field-service modules.

